Issue #131, opened 2019-02-22
PHP friendlyTime method is not working in Production environment and localhost


lablnet:
**Describe the bug**
when we get the time from database its treated as string so this line cause the problem
```php 
$time = (is_string($time)) ? strtotime($time) : $time;
```
https://github.com/zestframework/Zest_Framework/blob/master/src/Common/Formats.php#L88

Same problem in localHost
**Requirements**
1. Framework version: 3.0.0-dev
2. PHP version: 7..3
4. Enviroment: Hostinger shared hosting + windows xampp localhost
5. Error-log: none

lablnet:
@Maikuolan  could you help to solve this issue?

Maikuolan:
If `$time` can be an integer or a string, the corresponding annotation `@param (int)` should probably be `@param (int|string)` too, because `@param (int)` implies that the method should only accept integers.

Do we need to check whether the input is a string or an integer? I.e., is it possible that this information, when retrieved from the database, could be an integer or a string (i.e., two different possibilities)? Or, should this information in the database be consistent (i.e., always the same type of data)? I ask this, because there is a possibility of refactoring these methods, and maybe changing the logic a little.

If it's always an integer, the method becomes redundant anyway, and probably could be safely removed, I think. If it's always a string, `strtotime()` could be called directly without worrying about checking the type of data first. But, if it's sometimes an integer and sometimes a string, the check will still be necessary. Generally though, a database design that enforces consistent data types would be easier to maintain in the long term, I think.

lablnet:
@Maikuolan  
**Do we need to check whether the input is a string or an integer?**
yes because there is two posiblity like int or string
**is it possible that this information, when retrieved from the database, could be an integer or a string (i.e., two different possibilities)?**
yes

It may be string in case if user/developer do like this
```php
$time = "10 September 2000";
```
means to say English textual datetime.

But in case of int the time is actually timestamp so it will be int always, and timestamp will store in int type but i dont know why returning  as a string,  
i tried this issue fix when we use case (int) operator when calling this method like
https://github.com/zestframework/Zest_Web/commit/87d19a0ba13f1a13edd4fc99de1910210bd4f13b

and thanks for your suggestion 
https://github.com/zestframework/Zest_Framework/commit/b6e969a2477cc7717dc79f5ee38f3d591e0f6523

if there is better approach for condition check is string or int sire, you can send pull request or suggest here 

Thanks

Maikuolan:
[...]

PR #132, opened 2019-02-22
Fix for #131.

Maikuolan:


lablnet:
@Maikuolan  Thanks you so much sire

PR #155, opened 2019-03-05
Need Configuration file in component

lablnet:
Here json configuration file will be required in Component

**component,json**

structure like 

```json
{
    "name": "name",
    "id": "folder/dir_name",
    "version": "1.0.0",
    "zversion": "3.0.0",
    "status": true,
    "license": "license",
    "author": {
            "name": "name",
            "email":"email",
            "homepage": "website"
    }
}
```

### DIFF
--- a/src/Component/routes.php
+++ b/src/Component/routes.php
@@ -41,7 +41,6 @@ class routes extends Component
                 if ($config['status'] === true) {
                     require_once route()->com.$scans.'/routes.php';
                 }
-
             }
         }
         require_once 'dispatcher.php';
--- a/src/Component/routes.php
+++ b/src/Component/routes.php
@@ -33,13 +33,13 @@ class routes extends Component
         $path = '../App/Components/';
         $disk_scan = array_diff(scandir($path), ['..', '.']);
         foreach ($disk_scan as $scans) {
-            $configFile = route()->com . $scans . '/component.json';
+            $configFile = route()->com . $scans.'/component.json';
             if (file_exists($configFile)) {
                 $file = new FileHandling();
-                $c = $file->open($configFile,'readOnly')->read($configFile);
+                $c = $file->open($configFile, 'readOnly')->read($configFile);
                 $config = json_decode($c, true);
                 if ($config['status'] === true) {
-                    require_once route()->com .$scans.'/routes.php';
+                    require_once route()->com.$scans.'/routes.php';
                 }
 
             }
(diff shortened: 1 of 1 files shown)


src/Component/routes.php
```diff
@@ -31,7 +33,16 @@ public static function loadComs()
         $path = '../App/Components/';
         $disk_scan = array_diff(scandir($path), ['..', '.']);
         foreach ($disk_scan as $scans) {
-            require_once '../App/Components/'.$scans.'/routes.php';
+            $configFile = route()->com . $scans.'/component.json';
```
peter279k: If the configuration file is not existed, we should do something.

Maybe we should throw the exception about that?

lablnet: no as you see like 

````php
            if (file_exists($configFile)) {
                $file = new FileHandling();
                $c = $file->open($configFile,'readOnly')->read($configFile);
                $config = json_decode($c, true);
                if ($config['status'] === true) {
                    require_once route()->com .$scans.'/routes.php';
                }
            }
```

so we can check and ignore to include that component if there is no configuration file exists instead of showing expection

src/Component/routes.php
```diff
@@ -31,7 +33,15 @@ public static function loadComs()
         $path = '../App/Components/';
         $disk_scan = array_diff(scandir($path), ['..', '.']);
         foreach ($disk_scan as $scans) {
-            require_once '../App/Components/'.$scans.'/routes.php';
+            $configFile = route()->com.$scans.'/component.json';
```
peter279k, line 35: The `$disk_scan` and `$configFile` are the different approach of variables declaration.

And it should be concerned.

lablnet: yes, i think the second approach is best 
like ```configFile``` is that right?

peter279k: Yes. The second approach is better than first.

And other codes have same issue.

peter279k:
The codacy quality service is failed.

Could you check that before merging PR?

src/Component/routes.php
lablnet: yes thanks

no issue i dont know why this website show the issues
https://app.codacy.com/app/Lablnet/Zest_Framework/pullRequest?prid=3213058

lablnet:
@peter279k  so what you think

lablnet:
@peter279k  i deleted the app codacy

peter279k:
Hi @Lablnet, I think it's fine to do that I think.

At this moment, the `test` and `features` are more important than `code quality`.

lablnet:
Yes and also codacy says use of file_exists function is discourage like that useless errors i dont know why

Issue #166, opened 2019-03-09
if route not matched there is no expection


lablnet:
example: https://zestframework.xyz/index.php?dad

peter279k:
If the route is not matched, it means that the request URL is not found.

I think it should send the response the 404 HTTP status code.

lablnet:
fixed patched applied => https://zestframework.xyz/index.php?dad

PR #167, opened 2019-03-09
Proposal fix for  #166 

lablnet:
Proposal fix for  #166 
The issue with this proposal the Dispatch method become duplicate in
Zest Router class => https://github.com/zestframework/Zest_Framework/blob/master/src/Router/Router.php#L412
and in Component class {this pull request #167}

### DIFF
--- a/src/Component/Component.php
+++ b/src/Component/Component.php
@@ -19,6 +19,7 @@ namespace Zest\Component;
 use Zest\http\Request;
 use Zest\http\Response;
 use Zest\Input\Input;
+use Zest\View\View;
 
 class Component extends \Zest\Router\Router
 {
@@ -62,7 +63,7 @@ class Component extends \Zest\Router\Router
                 (is_object(isset($this->params['middleware']))) ? $this->params['middleware']->after(new Request(), new Response(), $this->params) : null;
             }
         } else {
-            throw new \Exception("No Route matched {$url}", 404);
+            View::view("errors/404",[],true,[],404);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


lablnet:
@peter279k  What do you think?

src/Component/Component.php
```diff
@@ -16,6 +16,53 @@
...
+                    $action = $this->params['action'];
+                    $action = $this->convertToCamelCase($action);
+                    if (preg_match('/action$/i', $action) == 0) {
+                        $controller_object->$action();
+                        (isset($this->params['middleware']) && is_object($this->params['middleware'])) ? (new $this->params['middleware']())->after(new Request(), new Response(), $this->params) : null;
+                    } else {
+                        throw new \Exception("Method $action in controller $controller cannot be called directly - remove the Action suffix to call this method");
+                    }
+                } else {
+                    throw new \Exception("Controller class $controller not found");
```
peter279k, line 58: If the route is not matched, it should send the `404` HTTP status code.

It means the request URL is not found.

peter279k: I don't think that throwing exception directly is a good way to solve this issue.

lablnet: yes, but what about if user want to customize this like in Laravel
http://laravel.com/dadasd

lablnet: oh @peter279k sorry for misunderstand we can fetch Views from View folder

peter279k: Just return view with customized error page :).

peter279k: Perhaps we can pass the default error page with view.

And let users customize their error pages.

lablnet: @peter279k  thanks

Issue #168, opened 2019-03-10
Response send method not working as it should be


lablnet:
Everything working fine but the body is not print out

https://github.com/zestframework/Zest_Framework/blob/master/src/http/Headers.php#L194

that was the same problem with our [website](https://zestframework.xyz/)

So in this case the view not working

https://github.com/zestframework/Zest_Framework/blob/master/src/View/View.php#L119

i revert this view to this, then the web works fine you can see above Zest framework link.
```php
public static function view($file, array $args = [], bool $minify = false, array $headers = [], $code = 200)
    {
        $headers['Content-Type'] = 'text/html';
        if ($minify === true) {
            $minify = new Minify();
            self::randerTemplate($file, $args);
            /*$config = [
                'code'    => $code,
                'body'    => $minify->htmlMinify(ob_get_clean(), 'code'),
                'headers' => [
                    $headers,
                ],
            ];
            $response = new Response($config);
            //$response->send();
              echo $response->send()*/
            echo $minify->htmlMinify(ob_get_clean(), 'code');
        } else {
            self::randerTemplate($file, $args);
            /*$config = [
                'code'    => $code,
                'body'    => ob_get_clean(),
                'headers' => [
                    $headers,
                ],
            ];
            $response = new Response($config);
            //$response->send();
              echo $response->send()*/
            echo ob_get_clean();
        }
    }
#reference https://github.com/zestframework/Zest_Framework/blob/master/src/View/View.php#L105
```

lablnet:
@peter279k  or @Maikuolan  looking for your help

peter279k:
It seems that the `Response` class should be the problem.

Consider the following `Headers::send` method:

```php
    public function send($code = null, array $headers = null)
    {
        if ($code !== null) {
            $this->withStatus($code);
        }
        if ($headers !== null) {
            $this->setHeaders($headers);
        }
        $body = $this->body;
        if (array_key_exists('Content-Encoding', $this->headers)) {
            $body = self::encodeBody($body, $this->headers['Content-Encoding']);
            $this->headers['Content-Length'] = strlen($body);
        }
        $this->sendHeaders();
        echo $body;
    }
```

I think the `Content-Encoding` header check should be the issue and I think it can remove them and try this code again.

lablnet:
Yes, you were right

following patch applied to  http://beta.zestframework.xyz/

```php
    public function send($code = null, array $headers = null)
    {
        if ($code !== null) {
            $this->withStatus($code);
        }
[...]

PR #171, opened 2019-03-12
fix for #168

lablnet:
fix proposal for #168

### DIFF
--- a/src/http/Headers.php
+++ b/src/http/Headers.php
@@ -194,7 +194,7 @@ abstract class Headers extends Clients\Client
             } elseif (preg_match('/ *deflate *,?/', $this->headers['accept-encoding'])) {
                 $encoding = 'deflate';
             }
-            (!empty($encoding)) ? $this->headers['Content-Encoding'] = $encoding : null;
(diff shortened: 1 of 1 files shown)


lablnet:
@peter279k  patched applied => http://beta.zestframework.xyz/

src/http/Headers.php
```diff
@@ -184,12 +185,18 @@ public function send($code = null, array $headers = null)
...
+        $this->headers['Content-Length'] = strlen($body);
+
+        if ($compress === true) {
+            $encoding = '';
+            if (preg_match('/ *gzip *,?/', $this->headers['accept-encoding'])) {
+                $encoding = 'gzip';
+            } elseif (preg_match('/ *deflate *,?/', $this->headers['accept-encoding'])) {
+                $encoding = 'deflate';
+            }
+            (!empty($encoding)) ? $this->headers['Content-Encoding'] = $encoding : null;
```
peter279k: Do you consider using the `isset`?

```php
(isset($encoding)) ? $this->headers['Content-Encoding'] = $encoding : null;
```

This will be more readable than `!empty($encoding)`.

lablnet: thanks, yes sure

PR #175, opened 2019-03-18
Avoid session hijacking

lablnet:


Maikuolan:
`session_regenerate_id(); //Aviod session hijacking`

( Aviod -> Avoid )

lablnet:
@Maikuolan  thanks for your response
You can help to solve this as soon as possible
`session_regenerate_id()` i think about to use this but i am not sure, where to use like when session starts /user login or destroy/logout/unset?

peter279k (commented):
Using the `session_regenerate_id` after the user sign out.

peter279k (approved):

lablnet:
Thanks @peter279k  and @Maikuolan

peter279k:
To avoid session hijacking, consider following code:

```php
if($_SERVER['REMOTE_ADDR'] !== $_SESSION['LAST_REMOTE_ADDR'] || $_SERVER['HTTP_USER_AGENT'] !== $_SESSION['LAST_USER_AGENT']) {
   session_destroy();
}
session_regenerate_id();
$_SESSION['LAST_REMOTE_ADDR'] = $_SERVER['REMOTE_ADDR'];
$_SESSION['LAST_USER_AGENT'] = $_SERVER['HTTP_USER_AGENT'];
```

The IP address is changed or `User-Agent` header is different.

We should be forced to let user sign out and destroy and regenerate session id.

Perhaps this will be an attacker try to get another session to try to do login work.

lablnet:
Yes so i think we need to store the user_agent and ip address in sessions

and in our `User` class we create method named `needRelogin()`
so we can used this method

Is i am right

PR #176, opened 2019-03-19
Added sitemap Package

lablnet:
## Usage
Just as an example
```php
<?php
namespace App\Controllers;

use Zest\Common\Sitemap\Sitemap;
use \Zest\Common\Sitemap\SitemapIndex;

class Sitemap extends \Zest\Controller\Controller
{

    public function index()
    {
        //Create sitenap
        $Sitemap = new Sitemap('malik.xml');
        $Sitemap->addItem('https://google.com',date('h:m:s'), 0.9, 'monthly');
        //Create sitemap index
        $SitemapIndex = new SitemapIndex('sitemap_index.xml');
        $SitemapIndex->AddItem('https://google.com/sitemap.xml');
    }
}

```

### DIFF
--- a/src/Common/Sitemap/Sitemap.php
+++ b/src/Common/Sitemap/Sitemap.php
@@ -1,0 +1,223 @@
+<?php
+
+/**
+ * This file is part of the Zest Framework.
+ *
+ * @author   Muhammad Umer Farooq <lablnet01@gmail.com>
+ * @author-profile https://www.facebook.com/Muhammadumerfarooq01/
+ *
+ * For the full copyright and license information, please view the LICENSE
+ *  file that was distributed with this source code.
+ *
+ * @since 3.0.0
+ *
+ * @license MIT
+ */
+
+namespace Zest\Common\Sitemap;
+
+class Sitemap extends SitemapWriter
+{
+    /**
+     * Start of sitemap.
+     *
+     * @since 3.0.0
+     *
+     * @var string
+     */
+    const START    = '<?xml version="1.0" encoding="UTF-8"?><urlset mlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd">';
+
+    /**
+     * End of sitemap.
+     *
+     * @since 3.0.0
+     *
+     * @var string
+     */    
+    const END     = '</urlset>';
+
+    /**
+     * Priority.
+     *
+     * @since 3.0.0
+     *
+     * @var array
+     */
+    private $validFrequencies = [
+        'always' , 'hourly', 'daily', 'weekly',
+        'monthly', 'yearly', 'never',
+    ];    
+
+    /**
+     * Priority.
+     *
+     * @since 3.0.0
+     *
+     * @var float
+     */
+    private $priority = 0.5;
+
+    /**
+     * Last modify.
+     *
+     * @since 3.0.0
+     *
+     * @var Datetime
+     */
+    private $lastMod;
+
+    /**
+     * Change frequency.
+     *
+     * @since 3.0.0
+     *
+     * @var string
+     */
+    private $changeFreq = 'weekly';
+
+    /**
+     * Sitemap file.
+     *
+     * @since 3.0.0
+     *
+     * @var string
+     */
+    private $file;
+
+    /**
+     * XML structure.
+     *
+     * @since 3.0.0
+     *
+     * @var string
+     */    
+    private $raw = '
+    <url>
+        <loc>:url</loc>
+        <lastmod>:lastmod</lastmod>
+        <changefreq>:changefreq</changefreq>
+        <priority>:priority</priority>
+    </url>';
+
+    /**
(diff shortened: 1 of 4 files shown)


src/Common/Sitemap/Sitemap.php
```diff
@@ -0,0 +1,223 @@
...
+    /**
+     * Determine whether the sitemap exists.
+     **
+     * @since 3.0.0
+     *
+     * @return bool
+    */
+    public function has()
+    {
+        return file_exists($this->file) ? true : false;
```
peter279k, line 204: This condition is not good because the `file_exists` should return bool value.

Please consider following code:

```php
......
return file_exists($this->file);
......
```

src/Common/Sitemap/Sitemap.php
```diff
@@ -0,0 +1,223 @@
...
+     * Delete the sitemap.
+     *
+     * @since 3.0.0
+     *
+     * @return object
+    */
+    public function delete()
+    {
+        if (file_exists($this->file)) {
+            unlink($this->$file);
```
peter279k, line 217: This `unlink` will be failed.

The `file_exists` should be true when the `$this->file` is directory.

It should use the `is_dir` and `file_exists` to check it's the file and it's existed.

src/Common/Sitemap/SitemapIndex.php
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * Determine whether the sitemap exists.
+     *
+     * @since 3.0.0
+     *
+     * @return bool
+    */
+    public function has()
+    {
+        return file_exists($this->file) ? true : false;
```
peter279k, line 161: It has the same issue I mention on previous comment.

src/Common/Sitemap/SitemapIndex.php
```diff
@@ -0,0 +1,180 @@
...
+     * Delete the sitemap.
+     *
+     * @since 3.0.0
+     *
+     * @return object
+    */
+    public function delete()
+    {
+        if (file_exists($this->file)) {
+            unlink($this->$file);
```
peter279k, line 174: It has the same issue I mention on previous comment.

src/Common/Sitemap/SitemapIndex.php
```diff
@@ -0,0 +1,180 @@
...
+    private function create($mode, $url, $lastMod = null)
+    {
+        $lastMod    = $lastMod     ?: $this->lastMod;
+        $raw = str_replace([':url', ':lastmod'], [$url, $lastMod], $this->raw);
+        if ($mode === 'create') {
+            $fileH = new SitemapWriter($this->file, 'writeOnly');
+            $fileH->write(SitemapIndex::START.PHP_EOL);
+            $fileH->write($raw);
+            $fileH->write(PHP_EOL.SitemapIndex::END);
+        } elseif ($mode === 'append') {
```
peter279k, line 106: I don't know what coding style is about the `elseif` condition.

We use the `else if` or `elseif` to be the consistency.

peter279k (commented):
I've added the request changes, please consider them.

Thanks.

src/Common/Sitemap/SitemapIndex.php
lablnet: could you explain more please?

src/Common/Sitemap/Sitemap.php
lablnet: Yes its return true so how this logics is wrong

i just test this 
```php
var_dump(file_exists(route()->public.'sitemap_index.xml') ? true : false);
```

Result if file exists => `bool(true)`
Result if file not exists => 'bool(false)`

src/Common/Sitemap/SitemapIndex.php
peter279k: I'm curious about the coding style.

If the `elseif` style is on all source code, it's fine to use this.

If not, please let these conditions be consistency.

src/Common/Sitemap/Sitemap.php
lablnet: No (Basically) the var `$this->file` is file with path like 

`C:\xampp\htdocs\dev\public/sitemap_index.xml`

src/Common/Sitemap/Sitemap.php
peter279k: That's right and the `?:` syntax is useless.

src/Common/Sitemap/SitemapIndex.php
lablnet: Oh i used `elseif` everywhere is this wrong???

I see this style also used by popular Frameworks like `Laravel`

src/Common/Sitemap/Sitemap.php
lablnet: so you means i can just return 

```php

return file_exists($this->file);
```
Okay thanks

src/Common/Sitemap/SitemapIndex.php
peter279k: That's okay.

This issue will be resolved.

src/Common/Sitemap/Sitemap.php
lablnet: https://github.com/zestframework/Zest_Framework/commit/c5f37051363570aeb98f255f6a4612dc03535602

Basically this pull was old and might be you saw code inconsity

PR #178, opened 2019-03-21
New proposal to get server/document root

lablnet:
But this not working at all as it should be, or i expected

### Change in App.php
```php
        'DATA_DIR'                => 'Data/',
        'CACHE_DIR'               => 'Cache/',
        'SESSION_PATH'            => 'Session/',
```

But following result show

Patched applied to 
http://beta.zestframework.xyz/


lablnet:
fixed

PR #183, opened 2019-03-24
Update .travis.yml

lablnet:


peter279k:
The following code is the problematic on `Configuration.php`:
```php
    /**
     * Prase the config file.
     *
     * @since 3.0.0
     *
     * @return array
     */
    public function parseData()
    {
        $data = [];
        $file1 = 'Config/App.php';
        $file2 = '../Config/App.php';
        if (file_exists($file1)) {
            $data += require $file1;
        } elseif (file_exists($file2)) {
            $data += require $file2;
        } else {
            throw new \Exception("Error, while loading Config {$file1} file", 404);
        }
        return $data;
    }

```

I think we should add the `$configPath` arg for this `parseData` method because it's should let developers be easy to specify their config file path.

If the developers don't specify file path, just try to detect the `$file1` and `$file2` by default.

BTW, why using the `+` operator to merge the arrays?

This will not a good way to manipulate arrays, and try using the `array_merge` instead.

This will be the another PR.

lablnet:
Yes the new issue is about https://github.com/zestframework/Zest_Framework/issues/185

**BTW, why using the + operator to merge the arrays?**
Crazy to say it concat the array, shorthand for the ```array_merge```

Yes i am writing Arrays class so it can be used their efficiently

PR #187, opened 2019-03-26
Create Arrays.php

lablnet:
Need to complete this.


lablnet:
@peter279k  Yes you're right i have removed implement `\ArrayAccess`

peter279k:
@Lablnet, actually, the `\ArrayAccess` is the interface.

And you have to implement these methods defined in this interface whatever you don't want to implement them.

If you want to keep implementing the `\ArrayAccess` interface, you can add these methods and keep them empty block or just don't let this `Arrays` class implement the `\ArrayAccess` interface.

peter279k (changes requested):
To keep the `Arrays` class simple and lightweight, we need to consider following approach:

- Remove the `\ArrayAccess` interface and the `Arrays` shouldn't implement this interface.
- Just add the `set`, `get`, `has`, `delete` (maybe) and `getValueByKey` for this `Arrays` class.

Here is the useful [link](https://github.com/selvinortiz/dot/blob/master/src/SelvinOrtiz/Dot/Dot.php) has the sample code about `Dot` array implementation with PHP.

lablnet:
@peter279k  thanks let check the latest commit

peter279k:
@Lablnet, could you squash all commits to one?

There are many commits on this PR.

src/Contracts/Data/Arrays,php
```diff
@@ -0,0 +1,129 @@
+<?php
```
peter279k: Wrong file name. It should be `src/Contracts/Data/Arrays.php`.

peter279k (commented):
It's the wrong file name and it causes the Travis CI build is failed.

lablnet:
Oh, I don't know how to could you do?, if possible

On Wednesday, March 27, 2019, peter279k <notifications@github.com> wrote:
> @Lablnet, could you squash all commits to one?
>
> There are many commits on this PR.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.<
https://ci4.googleusercontent.com/proxy/UZBzzCawb6jr3kxYqJxGv5lNEkXgV73T_W20C_TdsObyeq-lzTfky6z548vqKThKekB1CLQHhcFaJCWMEGfsuDmQvxNYZPlsQzdghB1VcwRy-IIo9aRwtEqgNMQP9BZHG6xqEZTX7j3MVqvWyotOg-W6lA=s0-d-e1-ft#https://github.com/notifications/beacon/AaeM2ThN1SLMp-MkUNebh-y0CIdMe5iIks5vaz4qgaJpZM4cKaK3.gif>

Issue #191, opened 2019-03-26
Fix Travis CI build


peter279k:
As title, the Travis CI build is failed due to the configuration file path.

Here is the [Travis CI build log](https://travis-ci.org/zestframework/Zest_Framework).

It's also related issue #185 and I think we can consider following approach to solve this:

- Create the `mock_config.php` file for do unit test in Travis CI build.

I think we can download the `config` file to the specify file path via `wget` command and this work can be defined in `.travis.yml` file.

lablnet:
Yes this work for know
But i am thinking about something like describe in https://github.com/zestframework/Zest_Framework/issues/185

So we can load configs file if exists and in our get `method` there will be option default value
so at core level we can pass default value, is this good approach what do you think?

i am curious How laravel manage their config and passsed tests?

peter279k:
In Laravel framework, it defines/initialize config in tests to do unit test for `Config` class.

You can consider the following code snippet in link:

[Config test](https://github.com/laravel/framework/blob/5.8/tests/Config/RepositoryTest.php)

lablnet:
I see, you're right 

the approach `Mock config` is right here

and you can do this by Sending `PR`

i am not sure but i think we can determine is code running from php unit testing
in `Config` class you can specify that condition to load config from mock file

peter279k:
@Lablnet. After tracing Laravel framework deeply, the `Hashing` work in `laravel/framework` doesn't load any config array.

I think you should separate the ArgonHasher to different the Hashers with different Argon2 algorithm.

You should look at the hashing implementation in [Laravel framework](https://github.com/laravel/framework/tree/5.8/src/Illuminate/Hashing).

We should avoid loading the config because it's not related to the core framework.

The config will only effect on web application and it's `zest/zest`.

The Laravel framework uses the `HashManager` and `HashServiceProvider` classes to load the hash algorithm driver name from config. and these classes are for `laravel/laravel`.

Perhaps we can scratch this approach. My approach is as follows:

- Separate the config loading work from related core `Hashing` classes.
- Using the `HashManager` or `HashServiceProvider` to load config work and dispatch the current Hashing class.
It's fine to ignore these classes in unit tests because they're the `Dispatcher`.
- We only work is to do unit tests for these core `Hashing` classes.

What do you think about that?

lablnet:
Yes you are absolutely right could you create a PR?

PR #192, opened 2019-03-26
Fix Hashing class test in Travis CI build

peter279k:
# Changed log
- Resolves issue #191.
- Fix `Hashing` class test.
- The `Hashing` and `Hash` classes shouldn't be tested because they're the `dispatcher` and `manager`.
They dispatch the current algorithm to do password hashing and the algorithm name is defined by `config` array.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .github
 /vendor
 /.phpintel
+/.phpunit.result.cache
--- a/Tests/Hashing/HashingTest.php
+++ b/Tests/Hashing/HashingTest.php
@@ -5,6 +5,7 @@ namespace Framework\Tests;
 namespace Zest\Tests\Hashing;
 
 use PHPUnit\Framework\TestCase;
+use Zest\Hashing\Argon2IdHashing;
 use Zest\Hashing\ArgonHashing;
 use Zest\Hashing\BcryptHashing;
 
@@ -26,16 +27,27 @@ class HashingTest extends TestCase
         if (!defined('PASSWORD_ARGON2I')) {
             $this->markTestSkipped('PHP not compiled with Argon2i hashing support.');
         }
-        if (!defined('PASSWORD_ARGON2ID')) {
-            $this->markTestSkipped('PHP not compiled with Argon2id hashing support.');
-        }
         $hashing = new ArgonHashing(['memory' => 512, 'time' => 5, 'threads' => 3]);
         $value = $hashing->make('password');
         $this->assertNotSame('password', $value);
         $this->assertTrue($hashing->verify('password', $value));
         $this->assertFalse($hashing->needsRehash($value));
-        $this->assertTrue($hashing->needsRehash($value, ['memory' => 512, 'time' => 5, 'threads' => 3]));
+        $this->assertFalse($hashing->needsRehash($value, ['memory' => 512, 'time' => 5, 'threads' => 3]));
         $this->assertSame('argon2i', password_get_info($value)['algoName']);
+    }
+
+    public function testBasicArgon2IdHashing()
+    {
+        if (!defined('PASSWORD_ARGON2ID')) {
+            $this->markTestSkipped('PHP not compiled with Argon2id hashing support.');
+        }
+        $hashing = new Argon2IdHashing(['memory' => 512, 'time' => 5, 'threads' => 3]);
+        $value = $hashing->make('password');
+        $this->assertNotSame('password', $value);
+        $this->assertTrue($hashing->verify('password', $value));
+        $this->assertFalse($hashing->needsRehash($value));
+        $this->assertFalse($hashing->needsRehash($value, ['memory' => 512, 'time' => 5, 'threads' => 3]));
+        $this->assertSame('argon2id', password_get_info($value)['algoName']);
     }
 
     public function testBasicBcryptVerification()
@@ -51,12 +63,20 @@ class HashingTest extends TestCase
         if (!defined('PASSWORD_ARGON2I')) {
             $this->markTestSkipped('PHP not compiled with Argon2i hashing support.');
         }
+        $original = 'password';
+        $hashing = new ArgonHashing(['memory' => 512, 'time' => 5, 'threads' => 3, 'verify' => true]);
+        $hashValue = $hashing->make($original);
+        $this->assertTrue($hashing->verify($original, $hashValue));
+    }
+
+    public function testBasicArgon2IdVerification()
+    {
         if (!defined('PASSWORD_ARGON2ID')) {
             $this->markTestSkipped('PHP not compiled with Argon2id hashing support.');
         }
         $original = 'password';
-        $hashing = new ArgonHashing(['memory' => 512, 'time' => 5, 'threads' => 3, 'verify' => true]);
+        $hashing = new Argon2IdHashing(['memory' => 512, 'time' => 5, 'threads' => 3, 'verify' => true]);
         $hashValue = $hashing->make($original);
-        $hashing->verify($original, $hashValue);
+        $this->assertTrue($hashing->verify($original, $hashValue));
     }
 }
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -5,6 +5,11 @@
             <directory>Tests</directory>
         </testsuite>
     </testsuites>
+    <filter>
+        <whitelist>
+            <directory suffix=".php">src</directory>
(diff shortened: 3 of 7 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 /.github
 .github
 /vendor
-/.phpintel
\ No newline at end of file
+/.phpintel
+/.phpunit.result.cache
```
peter279k, line 5: Sine the `PHPUnit 8 ` version, it will generate the cached result.

And the cached test result file should not be added in Git version control.

Tests/Hashing/HashingTest.php
```diff
@@ -51,12 +63,20 @@ public function testBasicArgon2iVerification()
...
+    {
         if (!defined('PASSWORD_ARGON2ID')) {
             $this->markTestSkipped('PHP not compiled with Argon2id hashing support.');
         }
         $original = 'password';
-        $hashing = new ArgonHashing(['memory' => 512, 'time' => 5, 'threads' => 3, 'verify' => true]);
+        $hashing = new Argon2IdHashing(['memory' => 512, 'time' => 5, 'threads' => 3, 'verify' => true]);
         $hashValue = $hashing->make($original);
-        $hashing->verify($original, $hashValue);
+        $this->assertTrue($hashing->verify($original, $hashValue));
```
peter279k, line 80: Add assertions to verify test for `Argon2IdHashing` class test.

phpunit.xml
```diff
@@ -5,6 +5,11 @@
             <directory>Tests</directory>
         </testsuite>
     </testsuites>
+    <filter>
+        <whitelist>
+            <directory suffix=".php">src</directory>
+        </whitelist>
+    </filter>
```
peter279k, line 12: To evaluate the code coverage, it should add the filter white list to achieve this.

src/Hashing/Argon2IdHashing.php
```diff
@@ -0,0 +1,249 @@
...
+ *  file that was distributed with this source code.
+ *
+ * @since 3.0.0
+ *
+ * @license MIT
+ */
+
+namespace Zest\Hashing;
+
+class Argon2IdHashing extends AbstractHashing
```
peter279k: This class is separated from `ArgonHashing` class because the original `ArgonHashing` class do two different algorithms.

They should be separated.

src/Hashing/ArgonHashing.php
```diff
@@ -232,20 +232,7 @@ public function getThreads()
      */
     protected function algorithm()
     {
-        $algo = __config()->hashing->driver;
```
peter279k, line 235: Just return the current algorithm name because the `Argon2` hashes have been separated.

src/Hashing/ArgonHashing.php
```diff
@@ -257,19 +244,6 @@ protected function algorithm()
      */
     protected function algorithmKeys()
     {
-        $algo = __config()->hashing->driver;
```
peter279k, line 260: It has the same reason that I mention in the previous comment.

src/Hashing/Argon2IdHashing.php
```diff
@@ -0,0 +1,249 @@
...
+    /**
+     * __construct.
+     *
+     * @param (array) $options
+     *
+     * @since 3.0.0
+     */
+    public function __construct(array $options)
+    {
+        $this->setMemory($options['memory'])->setTime($options['time'])->setThreads($options['threads']);
```
peter279k: I think it needs to check whether the keys are existed in specific `$options` array.

Consider following code snippets:

```php
$this->setMemory($options['memory'] ?? $this->memory)->setTime($options['time'] ?? $this->time)->setThreads($options['threads'] ?? $this->threads);
```

Should we need the `$options` validation? And I think it has the same issue on other codes.

Lablnet: Yes, better we can check, but i think these are required

lablnet: Yes , so i think we can  keep simple our setMemory() etc functions

lablnet:
Consider this code 
```php
<?php

/**
 * This file is part of the Zest Framework.
 *
 * @author   Muhammad Umer Farooq <lablnet01@gmail.com>
 * @author-profile https://www.facebook.com/Muhammadumerfarooq01/
 *
 * For the full copyright and license information, please view the LICENSE
 *  file that was distributed with this source code.
 *
 * @since 3.0.0
 *
 * @license MIT
 */

namespace Zest\Hashing;

class Argon2IdHashing extends ArgonHashing
{
    /**
     * Get the algroithm.
     *
     * @since 3.0.0
     *
     * @return \Constant
     */
    protected function algorithm()
    {
        return \PASSWORD_ARGON2ID;
    }

    /**
     * Get the algroithm keys.
     *
     * @since 3.0.0
     *
     * @return string
     */
    protected function algorithmKeys()
    {
        return 'argon2id';
    }
}
```

Instead of writing `Argon2IdHashing ` from scratch we can extends base `ArgonHashing` class to make it simple and powerful

what do you thimk

src/Hashing/Argon2IdHashing.php
peter279k: If we keep the related `set*` methods simple, we also have to consider the `BcryptHashing::setCost` method and its constructor.

Tests/Hashing/HashingTest.php
```diff
@@ -26,16 +27,27 @@ public function testBasicArgon2iHashing()
         if (!defined('PASSWORD_ARGON2I')) {
             $this->markTestSkipped('PHP not compiled with Argon2i hashing support.');
         }
+        $hashing = new ArgonHashing(['memory' => 512, 'time' => 5, 'threads' => 3]);
+        $value = $hashing->make('password');
+        $this->assertNotSame('password', $value);
+        $this->assertTrue($hashing->verify('password', $value));
+        $this->assertFalse($hashing->needsRehash($value));
+        $this->assertFalse($hashing->needsRehash($value, ['memory' => 512, 'time' => 5, 'threads' => 3]));
```
peter279k, line 35: This assertion is failed when check whether the hash needs hashing again.

The expected is false or true?

lablnet: false

lablnet: https://www.php.net/manual/en/function.password-needs-rehash.php

src/Hashing/Argon2IdHashing.php
lablnet: i have push commit to master branch for `BcryptHashing`

Tests/Hashing/HashingTest.php
peter279k: The assertions has been fixed.

src/Hashing/ArgonHashing.php
```diff
@@ -57,22 +57,22 @@ class ArgonHashing extends AbstractHashing
     /**
      * __construct.
      *
-     * @param (array) $options
+     * @param array $options
      *
      * @since 3.0.0
      */
-    public function __construct(array $options)
+    public function __construct(array $options = [])
```
peter279k: The `$options` should be the empty array by default, not null because the type hint is array.

src/Hashing/ArgonHashing.php
```diff
@@ -57,22 +57,22 @@ class ArgonHashing extends AbstractHashing
     /**
      * __construct.
      *
-     * @param (array) $options
+     * @param array $options
      *
      * @since 3.0.0
      */
-    public function __construct(array $options)
+    public function __construct(array $options = null)
```
peter279k, line 64: To keep the `$options` being null because we have to let `??` syntax be the expected behavior.

lablnet: yes you;re right

lablnet: you can fix so i can review PR

src/Hashing/ArgonHashing.php
lablnet: yes you're right

src/Hashing/ArgonHashing.php
peter279k: @Lablnet, I've fixed that and let `$options` be the `null` because of the `??` syntax.

lablnet (approved):

lablnet:
@peter279k  Thanks you so much for your grate contribution

PR #199, opened 2019-03-27
Create Arrays package

lablnet:


Tests/Data/ArraysTest.php
```diff
@@ -0,0 +1,133 @@
...
+use PHPUnit\Framework\TestCase;
+use Zest\Data\Arrays;
+
+class ArraysTest extends TestCase
+{
+    public function testIsReallyArray()
+    {
+        $this->assertTrue(Arrays::isReallyArray([0, 1]));
+        $this->assertTrue(Arrays::isReallyArray(['name' => 
+            'someName', 'username' => '@username']));
```
peter279k: This array should be constructed.

It looks like this don't have the proper coding style.

src/Data/Arrays.php
```diff
@@ -0,0 +1,439 @@
...
+     */
+    public static function has($array, $keys = null, $opr = null)
+    {
+        if (is_null($keys)) {
+            return false;
+        }
+
+        $keys = (array) $keys;
+
+        if ($keys === []) {
```
peter279k: How about using the  `count` to check whether the array length is zero?

src/Data/Arrays.php
```diff
@@ -0,0 +1,439 @@
...
+     *
+     * @param mixed $value Value to be checked.
+     *
+     * @since 3.0.0
+     *
+     * @return bool
+     */
+    public static function isReallyArray($value)
+    {
+        return is_array($value) && !empty($value);
```
peter279k: Please consider the following code snippet:

```php
return is_array($value) && count($value) !== 0;
```

This will be more clear than original.

The real array is to check whether the `$value` is array type and its length is not be zero.

src/Data/Arrays.php
```diff
@@ -0,0 +1,439 @@
...
+     * @param array|string $keys  Keys to search.
+     * @param string       $opr   Operator notaiton.
+     *
+     * @since 3.0.0
+     *
+     * @return bool
+     */
+    public static function has($array, $keys = null, $opr = null)
+    {
+        if (is_null($keys)) {
```
peter279k: It should use the `$keys === null` condition to check whether the `$keys` is null because of consistency.

src/Data/Arrays.php
```diff
@@ -0,0 +1,439 @@
...
+     *
+     * @see https://github.com/laravel/framework/blob/5.8/LICENSE.md
+     * @since 3.0.0
+     *
+     * @return void
+     */
+    public static function forget(&$array, $keys, $opr = null)
+    {
+        $arrOrg = $array;
+        if (is_null($keys)) {
```
peter279k: It has the same issue that I mention on previous comment.

src/Data/Arrays.php
```diff
@@ -0,0 +1,439 @@
...
+            return false;
+        }
+
+        $keys = (array) $keys;
+
+        if ($keys === []) {
```
peter279k: It has the same issue that I mention on previous comment.

src/Data/Arrays.php
```diff
@@ -0,0 +1,439 @@
...
+                // clean up before each pass
+                $array = &$original;
+
+                while (count($parts) > 1) {
+                    $part = array_shift($parts);
+
+                    if (isset($array[$part]) && is_array($array[$part])) {
+                        $array = &$array[$part];
+                    } else {
+                        continue 2;
```
peter279k, line 392: I'm not sure that `2` means.

peter279k: I know that. Here is the explanation:

 The continue 2 jumps to the next iteration of the loop at 2 levels back

peter279k (changes requested):
Some request changes need to be concerned.

- The StyleCI should be fixed, too.
- CI build is broken because there are some failed assertions and errors when PHPUnit do unit tests.
It should be fixed.

Tests/Data/ArraysTest.php
lablnet: What do you means, could you explain more please?

peter279k: It looks like the `'name' => 'someName'` is not at the same line.

Could you fix this?

lablnet:
The `Tests` here still failed, you can help to fix, because i tried and i dont understand why it failed

lablnet:
@peter279k Thanks you so much for your contribution

PR #200, opened 2019-03-29
Create EncryptionTest.php

lablnet:


lablnet:
Hello help to fix this issue, why test going fail?

peter279k:
@Lablnet, let me fetch this branch to look at detailed codes.

And it looks like you add other encryption approaches.

The new approaches seem to be failed in Travis CI build.

lablnet:
I fix one mistake

i was passing same `$str` to `decrypt` method

Tests/Encrption/EncryptionTest.php
```diff
@@ -0,0 +1,28 @@
...
+        $this->assertEquals($str, $openSslEncryption->decrypt($encryptHash));
+    }
+
+    public function testSodiumEncrypt()
+    {
+        $sodiumEncryption = new SodiumEncryption('123456key');
+        $str = 'This is a string';
+        $encryptHash = $sodiumEncryption->encrypt($str);
+        $this->assertNotEquals($str, $encryptHash);
+        $this->assertEquals($str, $sodiumEncryption->decrypt($encryptHash));
```
peter279k: Do you know the `assertEquals` and `assertSame`?

The `assertEquals` is about letting expected value be same as result value.

It's same as the `==` operator.

And their types will be different possibly.

The `assertSame` is about letting expected value be same as result value.

It's same as the `===` operator.

If we want to assert the expected value is equal to result, using the `assertEquals` is fine I think.

If we want to assert the expected value and type are same as result, using the `assertSame` should be proper I tihnk.

PR #201, opened 2019-03-29
Fix failed assertions for Arrays class test

peter279k:
# Changed log
- Resolves original PR #199.

### DIFF
--- a/src/Data/Arrays.php
+++ b/src/Data/Arrays.php
@@ -84,7 +84,7 @@ class Arrays implements ArraysContract
      * @param array  $array Array to be evaluated
      * @param string $key   Key
      * @param string $opr   Notation like 'dot'
-     * @param 
+     * @param
      *
      * @since 3.0.0
      *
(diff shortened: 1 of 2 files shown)


src/Data/Arrays.php
```diff
@@ -124,11 +124,14 @@ public static function set(&$array, $key = null, $value = null, $opr = null)
         }
 
         $keys = explode($opr, $key);
-        foreach ($keys as $ks => $key) {
-            if (!isset($array[$key])) {
-                $array = &$array[$key];
+        foreach ($keys as $key) {
+            if (!isset($array[$key]) || !is_array($array[$key])) {
```
peter279k, line 128: This refers the `Arr.php` class  in `illuminate/support` package.

src/Data/Arrays.php
```diff
@@ -363,10 +363,7 @@ public function subSetOfArray(array $array, $keys)
      */
     public static function forget(&$array, $keys, $opr = null)
     {
-        $arrOrg = $array;
-        if ($null !== $keys) {
-            return false;
-        }
```
peter279k, line 369: This should be useless.

When the `$keys` is null, it will be the empty `[]` array when casting to the array.

src/Data/Arrays.php
```diff
@@ -264,6 +262,8 @@ public static function multiToAssocWithSpecificOpr(array $arrays, $opr = null)
             if (self::isReallyArray($value) === true) {
                 $results = array_merge($results, self::multiToAssocWithSpecificOpr($value, $key.$opr));
             } else {
+                $key = $opr.$key;
+                $key = ($key[0] === '.' || $key[0] === '@' ? substr($key, 1) : $key);
```
peter279k, line 266: This is about fixing customized operators quickly.

Perhaps we should define an operator lists and using the `in_array` function to validate `$opr`.

Issue #194, opened 2019-03-27
Improvement in Container Class


lablnet:
Need improvement in Container class the current container class is damn

peter279k:
@Lablnet. Is this issue accomplished/resolved?

lablnet:
Yes i will open new issue regarding this

PR #203, opened 2019-03-30
Adding Container package

lablnet:
#194


src/Container/Container.php
```diff
@@ -0,0 +1,24 @@
+<?php
```
peter279k: The improved `Container` class need to have the unit tests.

We have to make sure that this `Container` class should be worked as our expected behavior.

lablnet: yes

lablnet:
@peter279k  check the latest commit and add `Tests` file in your available/free time

src/Container/Container.php
```diff
@@ -0,0 +1,465 @@
...
+	/**
+	 * Parse the hint parameter.
+	 *
+	 * @param  string|array $hint Type hint or array contaning both type hint and alias
+	 *
+	 * @since 3.0.0
+	 *
+	 * @return string
+	 */
+	protected function parseHint($hint): string
```
peter279k: It looks like the `$hint` will be the `array` or `string`.

Please check the type. I think it should throw `InvalidArgumentException` if `$hint` type is invalid :).

src/Container/Container.php
```diff
@@ -0,0 +1,465 @@
...
+	 * @param string|\Closure $class     Class name or closure.
+	 * @param bool            $singleton Should we return the same instance every time?
+	 *
+	 * @since 3.0.0
+	 *
+	 * @return void
+	 */
+	public function register($hint, $class, bool $singleton = false)
+	{
+		$this->hints[$this->parseHint($hint)] = ['class' => $class, 'singleton' => $singleton];
```
peter279k: Please check the `$hint` and `$class` type before assigning them to `$this->hints` array.

src/Container/Container.php
```diff
@@ -0,0 +1,465 @@
...
+	 * Creates a class instance using closure.
+	 *
+	 * @param closure $factory      Closuare.
+	 * @param array   $parameters Constructor parameters.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @return object
+	 */
+	public function closureFactory(\Closure $factory, $parameters)
```
peter279k: It should add the `array` type hint for `$parameters`.

src/Contracts/Container/Container.php
```diff
@@ -0,0 +1,156 @@
...
+     *
+     * @param string $class          Class.
+     * @param string $interface      Interface.
+     * @param string $implementation Implementation.
+     *
+     * @since 3.0.0
+     *
+     * @return void
+     */
+    public function registerContextualDependency($class, string $interface, string $implementation);
```
peter279k, line 69: I think it should add the `sting` type hint for `$class`.

src/Contracts/Container/Container.php
```diff
@@ -0,0 +1,156 @@
...
+     * Creates a class instance using closure.
+     *
+     * @param closure $factory      Closuare.
+     * @param array   $parameters Constructor parameters.
+     *
+     * @since 3.0.0
+     *
+     * @return object
+     */
+    public function closureFactory(\Closure $factory, $parameters);
```
peter279k: Add the `array` type hint for `$parameters`.

peter279k (changes requested):
Please consider the parameter type hints for all methods.

I write the comments on that line and note that considering type hints.

Some places that don't have comments, but it's similar problem on other comments.

You should check all argument type hints on methods :).

src/Container/Container.php
lablnet: Yes `$hint` will check in `parseHint` method
could you explain how we check the class parameter?

src/Container/Container.php
lablnet: Yes, consider the following code
```php
	protected function parseHint($hint): string
	{
		if(Arrays::isReallyArray($hint))
		{
			list($hint, $alias) = $hint;

			$this->aliases[$alias] = $hint;
		} elseif (!preg_match("/a-zA-Z/i", $hint)) {
			throw new \InvalidArgumentException ("The {$hint} parameter should be array or string ".gettype($hind)." given", 500);
			
		}

		return $hint;
	}
```

peter279k: @Lablnet, thank you for your reply.

This code snippets have following problems:

- coding style is about `if` condition should be consistency.
- I think we should use the `is_string` to validate the `$hint` is string before passing to the `preg_match`.

src/Container/Container.php
peter279k: I've mentioned some tips about type check on previous comment.

lablnet: so we can check like
```php
if (class_exists($class)) {
        //Do stuff....
}
```

src/Contracts/Container/Container.php
lablnet: explain more please

src/Container/Container.php
peter279k: I think the type hints validation should be another method.

And it's responsible for validating type.

src/Contracts/Container/Container.php
peter279k: It can use the `class_exists` to validate `$class`.

src/Container/Container.php
Lablnet: Okay thanks for your response

```php
	protected function parseHint($hint): string
	{
		if (is_string($hint) && preg_match("/a-zA-Z/i", $hint)) {
			return $hint;
		} elseif(Arrays::isReallyArray($hint)) {
			list($hint, $alias) = $hint;

			$this->aliases[$alias] = $hint;
		} else {
			throw new \InvalidArgumentException ("The {$hint} parameter should be array or string ".gettype($hind)." given", 500);
		}

		return $hint;
	}
```

lablnet: @peter279k  Thanks for your response, finally it become
```php
	protected function parseHint($hint): string
	{
		if (is_string($hint) && preg_match("/a-zA-Z/i", $hint)) {
			return $hint;
		} elseif(Arrays::isReallyArray($hint)) {
			list($hint, $alias) = $hint;
			$this->aliases[$alias] = $hint;

			return $hint;
		}

		throw new \InvalidArgumentException ("The {$hint} parameter should be array or string ".gettype($hind)." given", 500);
	}
```

src/Container/Container.php
lablnet: You mean it validate is type hint is string or array?

src/Container/Container.php
peter279k: Fix typo: `.gettype($hind)`. It's `$hint` :).

src/Container/Container.php
lablnet: could you explain more, i thought you mean we can append typeHint to parameter array

peter279k: Consider following code snippets:

```php
...
public function closureFactory(\Closure $factory, array $parameters)
...
```

src/Container/Container.php
lablnet: About "class_exists" actually its an object not class 
like we register as follow
```php
$this->container = new Container();
$this->container->registerSingleton([Arrays::class, 'array'], new \Zest\Data\Arrays);
```
so we cant able either class exists or not

peter279k: How about using the `instanceof`?

peter279k: Maybe the `class_exists` is fine, too.

lablnet:
@peter279k  Thanks you so much for your time to complete this class...

PR #206, opened 2019-03-30
Added Time Package

lablnet:


src/Contracts/Time/Time.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @license MIT
+ */
+
+namespace Zest\Contracts\Time;
+
+interface Time
+{
+
+	public function now();
```
peter279k: What kinds of time manipulation methods/works should be implemented?

Can you organize and define them?

lablnet: Updated you may see now.

src/Time/Time.php
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Determine whether the year is leap or not
+     *
+     * @param int $year A numeric year.
+     *
+     * @since 3.0.0
+     *
+     * @return  bool
+     */ 
+    public static function isLeapYear($year = null)
```
peter279k, line 64: I think we can let the `$year` be the `date('Y')` for the default value.

src/Time/Time.php
```diff
@@ -0,0 +1,133 @@
...
+    {
+        if (null === $year) {
```
peter279k: If the changes are accepted that I mention on previous comment.

This `null` condition can be removed.

src/Time/Time.php
```diff
@@ -0,0 +1,133 @@
...
+            $year = date('y');
+        }
+       if ((($year % 4 == 0) && ($year % 100 != 0)) || ($year % 400 == 0)) {
```
peter279k: This is the main condition for checking the specific year is leap.

And this should change into the one line:

```php
return ($year % 4 == 0 && $year % 100 != 0) || ($year % 400 == 0);
```

peter279k (changes requested):

lablnet:
@peter279k  review again and write test in your available time, in `ago` method we use placeholder technique to avoid language string ...

src/Contracts/Time/Time.php
```diff
@@ -0,0 +1,87 @@
...
+     * @since 3.0.0
+     *
+     * @return int
+     */
+    public static function daysInMonth($month = 1, $year = 1970);
+
+    /**
+     * Determine whether the year is leap or not.
+     *
+     * @param int $year A numeric year.
```
peter279k: The parameter will be the integer or null because the default value is `null`.

src/Contracts/Time/Time.php
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Converts a timestamp to GMT.
+     *
+     * @param int $time Unix timestamp
+     *
+     * @since 3.0.0
+     *
+     * @return int
+     */
+    public static function timestampToGmt($time = null);
```
peter279k, line 64: It has the same issue on the previous comment.

src/Contracts/Time/Time.php
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+namespace Zest\Contracts\Time;
+
+interface Time
+{
+    /**
+     * Get "now" time.
+     *
+     * @param string $timezone Valid php supported timezone.
```
peter279k: parameter will be `string` or `null`.

src/Contracts/Time/Time.php
```diff
@@ -0,0 +1,87 @@
...
+     * @since 3.0.0
+     *
+     * @return int
+     */
+    public static function timestampToGmt($time = null);
+
+    /**
+     * Converts the timestamp in to ago form.
+     *
+     * @param (int|string) $time Timestamp or English textual datetime (http://php.net/manual/en/function.strtotime.php)
```
peter279k: Remove `(` and `)`, please.

src/Contracts/Time/Time.php
```diff
@@ -0,0 +1,87 @@
...
+     * @since 3.0.0
+     *
+     * @return mixed
+     */
+    public static function ago($time);
+
+    /**
+     * Converts the timestamp in to h:m:s form.
+     *
+     * @param (int) $time Timestamp
```
peter279k: Remove `(` and `)`, please.

src/Time/Time.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Determine whether the year is leap or not.
+     *
+     * @param int $year A numeric year.
+     *
+     * @since 3.0.0
+     *
+     * @return bool
+     */
+    public static function isLeapYear($year = null)
```
peter279k, line 64: It has the same issue that mentions on previous comment.

src/Time/Time.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Converts a timestamp to GMT.
+     *
+     * @param int $time Unix timestamp
+     *
+     * @since 3.0.0
+     *
+     * @return int
+     */
+    public static function timestampToGmt($time = null)
```
peter279k, line 80: It has the same issue that mentions on previous comment.

src/Time/Time.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Converts the timestamp in to ago form.
+     *
+     * @param int|string $time Timestamp or English textual datetime (http://php.net/manual/en/function.strtotime.php)
+     *
+     * @since 3.0.0
+     *
+     * @return mixed
+     */
+    public static function ago($time)
```
peter279k:
```php
function time_elapsed_string($datetime, $full = false) {
    $now = new DateTime;
    $ago = new DateTime($datetime);
    $diff = $now->diff($ago);

    $diff->w = floor($diff->d / 7);
    $diff->d -= $diff->w * 7;

    $string = array(
        'y' => 'year',
        'm' => 'month',
        'w' => 'week',
        'd' => 'day',
        'h' => 'hour',
        'i' => 'minute',
        's' => 'second',
    );
    foreach ($string as $k => &$v) {
        if ($diff->$k) {
            $v = $diff->$k . ' ' . $v . ($diff->$k > 1 ? 's' : '');
        } else {
            unset($string[$k]);
        }
    }

    if (!$full) $string = array_slice($string, 0, 1);
    return $string ? implode(', ', $string) . ' ago' : 'just now';
}
```

To keep the time `ago` simple/lightweight, consider this function.

This approach is about using the [DateTime::diff](https://www.php.net/manual/en/datetime.diff.php) to get the  difference time between two date time.

We can also use the `DateTime` OO style or `date_create` to complete the difference work :).

Tests/Time/TimeTest.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * @dataProvider formatsSecondsDataProvider
+     */
+    public function testFormatsSecondsDataProvider($seconds, $expected)
+    {
+        $this->assertSame($expected, Time::formatsSeconds($seconds, 'Asia/Taipei'));
+    }
+}
```
peter279k, line 56: It needs more tests.

For example, the `ago`, `formatsSeconds` and so on.

peter279k (changes requested):
Some stuffs need to be concerned and enhanced.

Tests/Time/TimeTest.php
lablnet: yes you can send PR

Lablnet: i see but i think it will not work with timestamp

src/Time/Time.php
lablnet: i see but i think it will not work with timestamp

src/Time/Time.php
```diff
@@ -0,0 +1,158 @@
...
+            $ago = (round($time / $month) === 1) ? '1 :year' : round($time / $month).' :years';
+        }
+
+        return $ago;
+    }
+
+    /**
+     * Formats the time.
+     *
+     * @param (int|string) $time Timestamp or English textual datetime (http://php.net/manual/en/function.strtotime.php)
```
peter279k, line 154: Please remove the `(` and `)`.

src/Time/Time.php
lablnet: i see it will be work if we consider following code lets see
```php
    /**
     * Converts a timestamp to DateTime.
     *
     * @param int|null $time Unix timestamp.
     *
     * @since 3.0.0
     *
     * @return string
     */
    private static function timestampToDate($time = null)
    {
        $time = $time ?? time();
        $time = self::formatTime($time);
        $dateTime = new \DateTime();
        $dateTime->setTimestamp($time);

        return $dateTime->format('d-m-Y H:i:s');
    }

    /**
     * Converts the timestamp in to ago form.
     *
     * @param int|string $time Timestamp or English textual datetime (http://php.net/manual/en/function.strtotime.php)
     *
     * @author https://github.com/peter279k (https://github.com/zestframework/Zest_Framework/pull/206).
     * @since 3.0.0
     *
     * @return mixed
     */
    public static function ago($dateTime, $full = false)
    {
        $dateTime = self::timestampToDate($dateTime);
        $now = new \DateTime;
        $ago = new \DateTime($dateTime);
        $diff = $now->diff($ago);

        $diff->w = floor($diff->d / 7);
        $diff->d -= $diff->w * 7;

        $string = array(
            'y' => ':year',
            'm' => ':month',
            'w' => ':week',
            'd' => ':day',
            'h' => ':hour',
            'i' => ':minute',
            's' => ':second',
        );
        foreach ($string as $k => &$v) {
            if ($diff->$k) {
                $v = $diff->$k . ' ' . $v . ($diff->$k > 1 ? 's' : '');
            } else {
                unset($string[$k]);
            }
        }

        if (!$full) $string = array_slice($string, 0, 1);

        return $string ? implode(', ', $string) . ' ' : ':just';
    }
```

lablnet: FormatTime method
```php
        $time = preg_match('~\D~', $time) ? strtotime($time) : $time;
        return $time;
```

Tests/Time/TimeTest.php
peter279k: @Lablnet, thank you for your suggestion.

I think we can create another issue about enhancing the `Time` class unit tests after this PR is merged.

This PR I think we're focus on implementing  the `Time` classes completely.

lablnet: Yes its good grate thanks

src/Time/Time.php
peter279k: It looks good to me at this moment (without unit test).

lablnet: okay i am going to implement later we can fix iif any issue

lablnet:
@peter279k  Review the final so i can squash and merge

src/Time/Time.php
```diff
@@ -0,0 +1,182 @@
...
+     *
+     * @since 3.0.0
+     * @author https://github.com/Maikuolan (https://github.com/zestframework/Zest_Framework/issues/131)
+     *
+     * @return int
+     */
+    protected static function formatTime($time)
+    {
+        $time = preg_match('~\D~', $time) ? strtotime($time) : $time;
+        return $time;
```
peter279k, line 165: It should have the  empty new line before `return $time` because of consistency.

src/Time/Time.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        $dateTime = self::timestampToDate($dateTime);
+        $now = new \DateTime;
+        $ago = new \DateTime($dateTime);
+        $diff = $now->diff($ago);
+
+        $diff->w = floor($diff->d / 7);
+        $diff->d -= $diff->w * 7;
+
+        $string = array(
```
peter279k: Please use the `[]` short array syntax because of consistency.

peter279k (changes requested):
Most of works look good to me.

Some small request changes should be concerned :).

src/Time/Time.php
```diff
@@ -0,0 +1,182 @@
...
+     * @since 3.0.0
+     *
+     * @return string
+     */
+    public static function timestampToGmt($time = null)
+    {
+        $time = $time ?? time();
+        $dateTime = new \DateTime();
+        $dateTime->setTimestamp($time)->modify('+2 hours');
+        return $dateTime->format('d/m/Y H:i:s');
```
peter279k, line 86: It has the empty new line before `return` syntax because of consistency.

lablnet:
@peter279k Thanks you for your efforts and contribution

peter279k:
@Lablnet, it should have some issues we need to be concerned after this PR is merged.

- Fix coding style via StyleCI service.
- Create the issue about enhancing `Time` class unit test.

Issue #209, opened 2019-03-31
Scrutinizer error


Maikuolan:
Currently seeing this message from scrutinizer:

![composer](https://user-images.githubusercontent.com/12571108/55284191-214dce80-53a4-11e9-8f3f-8668be8444cd.png)

Not my area of expertise, so not something I could fix so easily, I think. Probably need to look more closely either at Composer package settings or at the unit tests, I think.

peter279k:
@Maikuolan, thank you for your reply.

The Scrutinizer CI error due to the PHPUnit dependencies issue.

And I think we have the following suggestion to solve this:

- Let this package be `php:7.1` at least and define this about the `require` block in `composer.json`.
- Define the PHPUnit version to be `phpunit: "^7.0 || ^8.0"` about `require-dev` block so that this dependencies issue will be solved.

lablnet:
@Maikuolan  thanks you so much for reporting the issue & @peter279k  thanks you for resolving..

peter279k:
@Lablnet, do we need to support the `php-7.1` version?

Or require `php-7.2` version at least?

lablnet:
`7.2`

lablnet:
Because we use phpUnit version 8

Maikuolan:
Sounds good. :-)

PR #210, opened 2019-03-31
Resolves issue #209

peter279k:
# Changed log

- Resolves issue #209.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,6 +3,9 @@
     "description": "The Zest Framework.",
     "keywords": ["framework", "zest"],
     "license": "MIT",
+    "require": {
+        "php": "^7.2"
+    },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -3,6 +3,9 @@
     "description": "The Zest Framework.",
     "keywords": ["framework", "zest"],
     "license": "MIT",
+    "require": {
+        "php": "^7.2"
```
peter279k, line 7: According to the `.travis.yml` setting, I think this framework require `php-7.2` version at least.

And we need to define the `php-7.2` version in `composer.json`.

lablnet: Yes you are right

lablnet (approved):

lablnet:
@peter279k Thanks you so much

PR #211, opened 2019-03-31
Adding & upgrading conversion package

lablnet:
Need to work on this too


Tests/Data/Conversion.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Framework\Tests;
+
+use PHPUnit\Framework\TestCase;
+use Zest\Data\Conversion;
+
+class ConversionTest extends TestCase
+{
+	public function testBitToBytes()
```
peter279k: The space should be 4 white spaces, not tab.

This should be fixed.

src/Data/Conversion.php
```diff
@@ -74,6 +76,49 @@ public static function objectArray($object)
...
+        $suffix = Arrays::arrayChangeCaseValue(['b', 'k', 'm', 'g', 't', 'p', 'e', 'z', 'y'], CASE_UPPER);
+        $f_base = (floor($base) > 8) ? 8 : floor($base);
+
+        return round(pow(1024, $base - floor($base)), $pre).$suffix[$f_base];
+    }
+
+    /**
+     * Convert the views to relative unit.
+     *
+     * @param int $size The value that you want provided
```
peter279k: The comment annotation is different from parameters inside method.

The `$size` is presented in comment annotation and the first parameter is `$n` in this method.

peter279k (changes requested):
These request changes should be concerned.

src/Contracts/Data/Conversion.php
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Convert arrays to Object.
+     *
+     * @param array $array Arrays
+     *
+     * @since 2.0.0
+     *
+     * @return object
+     */
+    public static function arrayObject($array);
```
peter279k: To be consistency of `function naming`, we should use the `arrayToObject` method.

lablnet: Yes you're right

src/Contracts/Data/Conversion.php
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Convert Objects to arrays.
+     *
+     * @param object $object
+     *
+     * @since 2.0.0
+     *
+     * @return array
+     */
+    public static function objectArray($object);
```
peter279k: The method name should change  into `objectToArray` because of consistency.

src/Contracts/Data/Conversion.php
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Convert XML to arrays.
+     *
+     * @param xml object $xml xml
+     *
+     * @since 2.0.0
+     *
+     * @return array
+     */
+    public static function xmlArray($xml);
```
peter279k: The method name should change into `xmlToArray` because of consistency.

src/Data/Conversion.php
```diff
@@ -29,10 +31,10 @@ class Conversion
      */
     public static function arrayObject($array)
```
peter279k: The method name should change into `arrayToObject` because of consistency.

src/Data/Conversion.php
```diff
@@ -74,6 +76,49 @@ public static function objectArray($object)
...
+        $f_base = (floor($base) > 8) ? 8 : floor($base);
+
+        return round(pow(1024, $base - floor($base)), $pre).$suffix[$f_base];
+    }
+
+    /**
+     * Convert the views to relative unit.
+     *
+     * @param int $n   Views.
+     * @param int $sep Seperator.
```
peter279k: This is weird.

The `$sep` is `integer` type annotation, but the default value is `,` defined in `views` method arguments.

You should consider that.

I think this variable should be `string`, not `int` :).

src/Contracts/Data/Conversion.php
```diff
@@ -0,0 +1,91 @@
...
+     * @since 3.0.0
+     *
+     * @return mixed
+     */
+    public static function views($n, $sep = ',');
+
+    /**
+     * Convert XML to arrays.
+     *
+     * @param xml object $xml xml
```
peter279k: The `xml object` annotation?

I'm not sure about what this type is defined here.

Tests/Data/Conversion.php
```diff
@@ -0,0 +1,17 @@
...
+class ConversionTest extends TestCase
+{
+    public function testBitToBytes()
+    {
+        $this->assertSame('1K', Conversion::bitToBytes(1024));
+        $this->assertNotSame('1K', Conversion::bitToBytes(1024));
+        $this->assertSame('9.31G', Conversion::bitToBytes(1024));
+        $this->assertNotSame('9G', Conversion::bitToBytes(1024));
+    }
+}
```
peter279k, line 15: This should have more unit tests on this `ConversionTest` class test :).

src/Contracts/Data/Conversion.php
lablnet: I think mixed is best here...

PR #212, opened 2019-03-31
The sodium key should be 32

lablnet:


peter279k:
This is about the quick fix for Sodium key size.

And we consider the following code snippets in `SodiumEncryption` class:

```php
....
public function __construct($key = null)
    {
        if (!function_exists('sodium_crypto_secretbox_keygen')) {
            throw new \Exception('The sodium php extension does not installed or enabled', 500);
        }

        if (null !== $key) {

            //Should use user define key.
            $this->key = substr(hash('sha512', $key), 0, 32);
        } else {
            throw new \Exception('Crypto key not found', 500);
        }
    }
......
```

This will not validate the `$key` size length in this class constructor.

We need to do this validation to mention the developers the key size is invalid.

And key size should be `32`.

lablnet:
@peter279k  thanks, look at this => https://github.com/zestframework/Zest_Framework/pull/213

PR #213, opened 2019-03-31
Update SodiumEncryption.php

lablnet:


src/Encryption/Adapter/SodiumEncryption.php
```diff
@@ -30,9 +30,13 @@ public function __construct($key = null)
         }
 
         if (null !== $key) {
-
-            //Should use user define key.
-            $this->key = substr(hash('sha512', $key), 0, 32);
+            //Need to validate key to o mention the developers the key size is invalid. [@peter279k(https://github.com/peter279k)]
+            if (strlen($key) >= 32) {
```
peter279k: This condition is not good.

How about considering following approach:

- Validate the `$key` length is `32`. We can use the `if (strlen($key) !== 32) {.....`.
- If not, throw the `InvalidArgumentException`.

peter279k (changes requested):

src/Encryption/Adapter/SodiumEncryption.php
```diff
@@ -30,9 +30,13 @@ public function __construct($key = null)
         }
 
         if (null !== $key) {
-
-            //Should use user define key.
-            $this->key = substr(hash('sha512', $key), 0, 32);
+            //Need to validate key to o mention the developers the key size is invalid. [@peter279k(https://github.com/peter279k)]
+            if (strlen($key) !== 32) {
+                throw new \InvalidArgumentException('The key size is invalid, and key size should be 32', 500);
+            } else {
```
peter279k, line 36: This `else` condition should be removed.

When throwing the `InvalidArgumentException` exception, the program will be terminated.

PR #214, opened 2019-04-04
Add now, daysInMonth. Enhance isLeapYear

peter279k:
# Changed log
- Implement the `now` and `daysInMonth` methods.
- Enhance `isLeapYear` and `formatsSeconds` methods.

### DIFF
--- a/src/Common/TimeZone.php
+++ b/src/Common/TimeZone.php
@@ -21,7 +21,7 @@ class TimeZone extends \DateTimeZone
     /**
      * Set Default timeZone.
      *
-     * @param (string) $tz valid time zone
+     * @param string $tz valid time zone
      *
      * @since 3.0.0
      *
@@ -55,6 +55,6 @@ class TimeZone extends \DateTimeZone
      */
     public static function validate($tz) :string
     {
-        return (in_array($tz, static::tZIdentifiers()) === true) ? $tz : 'UTC';
+        return in_array($tz, static::tZIdentifiers()) ? $tz : 'UTC';
     }
 }
--- a/src/Time/Time.php
+++ b/src/Time/Time.php
@@ -16,78 +16,75 @@
 
 namespace Zest\Time;
 
+use Zest\Common\TimeZone;
 use Zest\Contracts\Time\Time as TimeContract;
 
 class Time implements TimeContract
 {
     /**
-     * Get "now" time
+     * Get "now" time.
      *
      * @param string $timezone Valid php supported timezone.
      *
      * @since 3.0.0
      *
-     * @return  int
+     * @return int
      */
-    public static function now($timezone = NULL)
+    public static function now($timezone = '')
     {
+        TimeZone::seteDefaultTz($timezone);
 
+        return time();
     }
 
     /**
-     * Number of days in a month
+     * Number of days in a month.
      *
      * @param int $month A numeric month.
      * @param int $year  A numeric year.
      *
      * @since 3.0.0
      *
-     * @return  int
+     * @return int
      */
     public static function daysInMonth($month = 1, $year = 1970)
     {
-
+        return cal_days_in_month(CAL_GREGORIAN, $month, $year);
     }
 
     /**
-     * Determine whether the year is leap or not
+     * Determine whether the year is leap or not.
      *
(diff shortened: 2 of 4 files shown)


src/Common/TimeZone.php
```diff
@@ -21,7 +21,7 @@ class TimeZone extends \DateTimeZone
     /**
      * Set Default timeZone.
      *
-     * @param (string) $tz valid time zone
```
peter279k, line 24: Follow the PHPDoc comment annotation style and remove the `()`.

src/Common/TimeZone.php
```diff
@@ -55,6 +55,6 @@ public static function tZIdentifiers() :array
      */
     public static function validate($tz) :string
     {
-        return (in_array($tz, static::tZIdentifiers()) === true) ? $tz : 'UTC';
+        return in_array($tz, static::tZIdentifiers()) ? $tz : 'UTC';
```
peter279k, line 58: The `in_array` will return bool value, and removing the `=== true` statement makes sense.

src/Time/Time.php
```diff
@@ -16,118 +16,111 @@
...
      *
      * @param string $timezone Valid php supported timezone.
      *
      * @since 3.0.0
      *
-     * @return  int
+     * @return int
      */
-    public static function now($timezone = NULL)
+    public static function now($timezone = '')
```
peter279k, line 33: The `$timezone` will be the `string` type, not `null`.

And it makes sense that let the `$timezone` be empty string by default.

src/Time/Time.php
```diff
@@ -16,118 +16,111 @@
...
      *
      * @since 3.0.0
      *
-     * @return  int
+     * @return int
      */
     public static function daysInMonth($month = 1, $year = 1970)
     {
-
+        return cal_days_in_month(CAL_GREGORIAN, $month, $year);
```
peter279k, line 52: Using the `cal_days_in_month` function which is provided by `calendar` extension in PHP to calculate the days of given month.

src/Time/Time.php
```diff
@@ -16,118 +16,111 @@
...
+     */
     public static function isLeapYear($year = null)
     {
-        if (null === $year) {
-            $year = date('y');
-        }
-       if ((($year % 4 == 0) && ($year % 100 != 0)) || ($year % 400 == 0)) {
-            return true;
-       }
+        $year = $year ?? date('Y');
```
peter279k, line 66: Using the `??` syntax to check whether the `$year` is assigned.

src/Time/Time.php
```diff
@@ -16,118 +16,111 @@
...
 
-       return false;
+        return (($year % 4 == 0) && ($year % 100 != 0)) || ($year % 400 == 0);
```
peter279k, line 68: Remove the `if` condition. Using the `?:` syntax  instead.

src/Time/Time.php
```diff
@@ -16,118 +16,111 @@
...
-     * @param (int) $time Timestamp
+     * @param int    $time     Timestamp
+     * @param string $timezone Valid php supported timezone.
      *
      * @since 3.0.0
      *
      * @return mixed
      */
-    public static function formatsSeconds($seconds)
+    public static function formatsSeconds($seconds, $timezone = '')
```
peter279k, line 120: We should give the `$timezone` name to set timezone before formatting the seconds.

And the default timezone is `UTC` if the `$timezone` is empty.

src/Time/Time.php
```diff
@@ -16,118 +16,111 @@
...
     {
-        $h = ($seconds >= 3600) ? (int) round($seconds / 3600) : 0;
-        $time = ($seconds >= 3600) ? $seconds % 3600 : $seconds;
-        $m = (int) $time / 60;
-        $s = (int) $time % 60;
+        TimeZone::seteDefaultTz($timezone);
 
-        return (int) $h.':'.(int) $m.':'.(int) $s;
+        return date('h:i:s', $seconds);
```
peter279k, line 124: Using the native `date` function in PHP to format seconds.

I think this approach will be more efficient than using own second calculation.

lablnet:
@peter279k  thanks you so much

Issue #223, opened 2019-05-02
Enhancement of Arrays class


lablnet:
I think we should implement four more methods in our `Arrays` class
### 1. Remove duplicate
there should be method namely `removeDuplicates()` which remove duplicate records in array consider following data set
```php
        $dataSet = [
            'users' => 
            [
                'id' => 1,
                'name' => "Umer",
                'username' => 'peter',
            ],
            [
                'id' => 2,
                'name' => "Peter Khot",
                'username' => 'peter',
            ]
        ];
``` 
Look at above problem set and see the username become duplicate, we don't think how it happen, but there should be way/method which remove duplicate by specific key pair.

### 2. Pluck
To retrieves all of the values for a given key from an array
Consider following dataset 

```php
        $array = [
            ['developer' => ['id' => 1, 'name' => 'Umer']],
            ['developer' => ['id' => 2, 'name' => 'Peter']],
        ];
```
if we do something following ``` Arr:ays:pluck($array, 'developer.name');``` it should return both name.

### 3. Most Occurring
Like anyone wants return most occuring name from array so there should be method that works with every datatype to get most occurring values.
consider the following dataset
```php
        $dataSet = [
            'users' => 
            [
                'id' => 1,
                'name' => "Umer",
                'username' => 'peter',
            ],
            [
                'id' => 2,
                'name' => "Umer",
                'username' => 'umer01'
            ],
            [
                'id' => 3,
                'name' => "Peter Khot",
                'username' => 'peter',
            ]
        ];
```
so if we wants most occuring name it should return `Umer` [It should works with every data type even with bool]

### 4 Least Occurring 
Like anyone wants return least occuring name from array so there should be method that works with every datatype to get most occurring values.
consider the following dataset
```php
        $dataSet = [
            'users' => 
            [
                'id' => 1,
                'name' => "Umer",
                'username' => 'peter',
            ],
            [
                'id' => 2,
                'name' => "Umer",
                'username' => 'umer01'
            ],
            [
                'id' => 3,
                'name' => "Peter Khot",
                'username' => 'peter',
            ]
        ];
```
so if we wants least occuring name it should return `Peter Khot` [It should works with every data type even with bool]

**Here above all are real world problem sets**

What do you think @peter279k  and @Maikuolan

lablnet:
Now last `method` is left still 

### 2. Pluck
To retrieves all of the values for a given key from an array
Consider following dataset 

```php
        $array = [
[...]

PR #224, opened 2019-05-03
Enhancement of Arrays class #223

lablnet:
Enhancement of Arrays class #223

### DIFF
--- a/Tests/Data/ArraysTest.php
+++ b/Tests/Data/ArraysTest.php
@@ -207,6 +207,6 @@ class ArraysTest extends TestCase
            6 => 7,
            7 => 8,
            8 => 9
-        ], Arrays::removeDuplicates($dataSet3, 'username'));
+        ], Arrays::removeDuplicates($dataSet3));
     }
 }
--- a/Tests/Data/ArraysTest.php
+++ b/Tests/Data/ArraysTest.php
@@ -197,6 +197,16 @@ class ArraysTest extends TestCase
             ],
         ], Arrays::removeDuplicates($dataSet1, 'username'));
         $this->assertSame(['a' => 'green', 0 => 'red', 1 => 'blue' ], Arrays::removeDuplicates($dataSet2, 'username'));
-        $this->assertSame([1,2,3,4,5,6,7,8,9], Arrays::removeDuplicates($dataSet3, 'username'));
+        $this->assertSame([
+           0 => 1,
+           1 => 2,
+           2 => 3,
+           3 => 4,
+           4 => 5,
+           5 => 6,
+           6 => 7,
+           7 => 8,
+           8 => 9
+        ], Arrays::removeDuplicates($dataSet3, 'username'));
     }
 }
--- a/Tests/Data/ArraysTest.php
+++ b/Tests/Data/ArraysTest.php
@@ -174,7 +174,7 @@ class ArraysTest extends TestCase
                     [
                         'id'       => 2,
                         'name'     => 'Umer',
-                        'username' => 'umer01'
+                        'username' => 'umer01',
                     ],
                     [
                         'id'       => 3,
@@ -202,7 +202,7 @@ class ArraysTest extends TestCase
                 ['a' => 'green', 0 => 'red', 1 => 'blue'],
             ],
             [
-                [1,2,3,4,2,5,6,3,7,8,9],
+                [1, 2, 3, 4, 2, 5, 6, 3, 7, 8, 9],
                 '',
                 [
                     0  => 1,
@@ -213,7 +213,7 @@ class ArraysTest extends TestCase
                     6  => 6,
                     8  => 7,
                     9  => 8,
-                    10 => 9
+                    10 => 9,
                 ],
             ],
         ];
--- a/Tests/Data/ArraysTest.php
+++ b/Tests/Data/ArraysTest.php
@@ -182,7 +182,7 @@ class ArraysTest extends TestCase
             ]
         ];
         $dataSet2 = array("a" => "green", "red", "b" => "green", "blue", "red");
-
+        $dataSet3 = [1,2,3,4,2,5,6,3,7,8,9];
         $this->assertSame([
             0 => [
                 'id'       => 1,
@@ -196,5 +196,6 @@ class ArraysTest extends TestCase
             ],
         ], Arrays::removeDuplicates($dataSet1, 'username'));
         $this->assertSame(['a' => 'green', 0 => 'blue', 1 => 'red' ], Arrays::removeDuplicates($dataSet2, 'username'));
+        $this->assertSame([1,2,3,4,5,6,7,8,9], Arrays::removeDuplicates($dataSet3, 'username'));
     }
 }
--- a/Tests/Data/ArraysTest.php
+++ b/Tests/Data/ArraysTest.php
@@ -166,7 +166,7 @@ class ArraysTest extends TestCase
         return [
             [
                 [
-                    'users' => 
+                    'users' =>
                     [
                         'id'       => 1,
                         'name'     => "Umer",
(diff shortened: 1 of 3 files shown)


peter279k:
The assertions are failed in Travis CI build.

You should try to fix that.

Here is the failed assertion captured picture:

image

Tests/Data/ArraysTest.php
```diff
@@ -160,4 +160,42 @@ public function testArrayChangeCaseValue()
...
+                'name'     => "Umer",
+                'username' => 'umer01'
+            ],
+            [
+                'id'       => 3,
+                'name'     => "Peter Khot",
+                'username' => 'peter',
+            ]
+        ];
+        $dataSet2 = array("a" => "green", "red", "b" => "green", "blue", "red");
```
peter279k: Please use the short array syntax because of the consistency.

peter279k: It seems that the expected array will be changed the order in associative array.

peter279k:
> The assertions are failed in Travis CI build.
> 
> You should try to fix that.
> 
> Here is the failed assertion captured picture:
> 
> image

It seems that the expected associative array should change the order.

Tests/Data/ArraysTest.php
```diff
@@ -160,4 +160,53 @@ public function testArrayChangeCaseValue()
         $this->assertSame(['name' => 'ALEX'], Arrays::arrayChangeCaseValue($array, CASE_UPPER));
         $this->assertNotSame(['name' => 'Alex'], Arrays::arrayChangeCaseValue($array, CASE_UPPER));
     }
+
+    public function testRemoveDuplicates()
```
peter279k: Please consider the PHPUnit data provider feature to split the test cases from `testRemoveDuplicates` method.

Here is the reference:

https://phpunit.readthedocs.io/en/latest/writing-tests-for-phpunit.html?highlight=data%20provider#data-providers

peter279k (approved):

src/Data/Arrays.php
```diff
@@ -478,4 +478,42 @@ public static function arrayChangeCaseValue($array, $case = CASE_LOWER)
...
+            return array_unique($array);
+        }
+        if (self::isMulti($array) && empty($key)) {
+            return false;
+        }
+        $dataSet = [];
+        $i = 0;
+        $keys = [];
+        foreach ($array as $k) {
+            if (in_array($k[$key], $keys)) {
```
peter279k, line 507: Using the `in_array` is slower than `empty`.

Please consider following code snippets:

```php
......
if (empty($keys[$k[$key]])) {
......
```

src/Data/Arrays.php
```diff
@@ -478,4 +478,42 @@ public static function arrayChangeCaseValue($array, $case = CASE_LOWER)
...
+     * Remove duplicate values from array.
+     *
+     * @param array      $array The array to work on.
+     * @param string|int $key   Key that need to evaulate.
+     *
+     * @since 3.0.0
+     *
+     * @return array
+     */
+    public static function removeDuplicates(array $array,$key = '')
```
peter279k: I wonder why the specific `$key` has the default value.

I think the default value should be removed because the specific `$key` will be the required argument, not optional.

lablnet: Here key is optional arg, because key only required if working on Multidimensional arrays. in sequential or assoc array key is not required so i make it optional

Tests/Data/ArraysTest.php
lablnet: I looked but wonder, 
```php
public function test($expected) {
}
```
here how i know the values of `$expected` variable? if i don't know so how could i match that either method working fine or not.

peter279k: @Lablnet, the PHPUnit will tell you which test case numbers are failed during unit test execution.

And output the failed assertion messages.

The `test` method should have the multiple arguments that are up to data provider multidimensional array.

You can consider using the following code snippets:

```php
......
public function test($data, $expected)
{
// tests......
}
......
```

src/Data/Arrays.php
peter279k: @Lablnet, ok. It makes sense, but I just note that the key will be `index` or `key` name.

The key is required for multidimensional or ordered array.

lablnet: key will be `key` not index, as you may see code i use default php function for handling indexes and assoc arrays

src/Data/Arrays.php
lablnet: I have tested
```php
if (!empty($keys[$k[$key]])) {
```
its not working and i think it should be work but no idea why it does not works

peter279k: Could you provide the sample code that you test?

lablnet: Well

#### Method
```php
    public static function removeDuplicates(array $array,$key = '')
    {
        if (!self::isReallyArray($array)) {
            return false;
        }
        if (self::isSequential($array) || self::isAssoc($array)) {
            return array_unique($array);
        }
        if (self::isMulti($array) && empty($key)) {
            return false;
        }
        $dataSet = [];
        $i = 0;
        $keys = [];
        foreach ($array as $k) {
            if (!empty($keys[$k[$key]])) {
                continue;
            } else {
                $keys[$i] = $k[$key];
                $dataSet[$i] = $k;
            }

            $i++;
        }

        return $dataSet;
    }
```

### Usage
```php
        $dataSet = [
            'users' => 
            [
                'id' => 1,
                'name' => "Umer",
                'username' => 'peter',
            ],
            [
                'id' => 2,
                'name' => "Umer",
                'username' => 'umer01'
            ],
            [
                'id' => 2,
                'name' => "Peter Khot",
                'username' => 'peter',
            ]
        ];
        echo "<pre>";
        print_r(Arrays::removeDuplicates($dataSet,'username'));
        echo "</pre>";
```

The condition should be `!empty()` instead of `empty()` because if not empty then we want continue in order to avoid from duplicate records, but this still not works, it gives us back all records you may check yourself

Thanks @peter279k

peter279k: Ok. It makes sense.

Thank you for your explanation.

Tests/Data/ArraysTest.php
lablnet: could you provide sample example please @peter279k  i am still confused!!

peter279k: The sample code is as follows:

```suggestion

    public function removeDuplicatesDataProvider()
    {
        return [
            [
                [
                    'users' => 
                    [
                        'id'       => 1,
                        'name'     => "Umer",
                        'username' => 'peter',
                    ],
                    [
                        'id'       => 2,
                        'name'     => "Umer",
                        'username' => 'umer01'
                    ],
                    [
                        'id'       => 3,
                        'name'     => "Peter Khot",
                        'username' => 'peter',
                    ]
                ],
                'username',
                [
                    0 => [
                        'id'       => 1,
                        'name'     => 'Umer',
                        'username' => 'peter'
                    ],
                    1 => [
                        'id'       => 2,
                        'name'     => 'Umer',
                        'username' => 'umer01'
                    ],
                ],
            ],
            [
                ["a" => "green", "red", "b" => "green", "blue", "red"],
                'username',
                [
                    0 => [
                        'id'       => 1,
                        'name'     => 'Umer',
                        'username' => 'peter'
                    ],
                    1 => [
                        'id'       => 2,
                        'name'     => 'Umer',
                        'username' => 'umer01'
                    ],
                ],
            ],
            [
                [1,2,3,4,2,5,6,3,7,8,9],
                '',
                [
                    0 => 1,
                    1 => 2,
                    2 => 3,
                    3 => 4,
                    5 => 5,
                    6 => 6,
                    8 => 7,
                    9 => 8,
                    10 => 9
                ],
            ],
        ];
    }

    /**
     * @dataProvider removeDuplicatesDataProvider
     */
    public function testRemoveDuplicates($dataSet, $specificKey, $expected)
    {
        $this->assertSame($expected, Arrays::removeDuplicates($dataSet, $specificKey));
    }
```

As you can see, the tested data sets are split from `testRemoveDuplicates` method.

peter279k:
The Travis CI build is failed.

```
PHP Fatal error:  Class Zest\Data\Arrays contains 2 abstract methods and must therefore be declared abstract or implement the remaining methods 
```

The `Arrays` class should implement all methods which defined in specific interface.

lablnet:
@peter279k  see `Arrays` class all method defined

peter279k:
I think it is ready to be merged.

Before merging the PR, we should fix coding style.

peter279k (approved):

lablnet:
@peter279k  What about the `tests` that left and one `method` => `pluck` ?

peter279k:
> @peter279k What about the `tests` that left and one `method` => `pluck` ?

I think we can create another PR to enhance the Array test.

lablnet:
Okay sure let me finish the tests for `random`, `where`, `shuffle` and `query` method

PR #227, opened 2019-05-10
Fix failed assertion

peter279k:
# Changed log
- Fix failed assertion.
- It's related to PR #224.


lablnet:
@peter279k thanks you so much

PR #228, opened 2019-05-11
Fix failed testRandom assertion

peter279k:
# Changed log
- Fix `Arrays::random` method test.
- The first parameter is in `PHPUnit\Framework\TestCase::assertContains` method should be specific value.
The `Array::random` method will return the array type with one value and we have to get that value and pass to the first parameter in `assertContains` method.
- It's related to issue #223.


lablnet:
@peter279k  thanks

lablnet (approved):

Issue #234, opened 2019-05-22
Be flexible for the format inside Time class


peter279k:
**Is your feature request related to a problem? Please describe.**
I get started to think about the date format inside `Time` class.

It looks like the date format has the fixed format and developer doesn't have any option to make the date be formatted.

Such as `Time::timestampToGmt` result will be the `d/m/Y H:i:s` format.

**Describe the solution you'd like**
Perhaps we consider have the option to define the date format string.

**Possible approach**
I suggest the following approach and we should follow this rule.

- Add `setDateFormat` and `getDateFormat` methods and define private/protected `dateFormat` variable.
- Add the `validateDateFormat` method when setting `setDateFormat`.

PR #237, opened 2019-08-01
Enhancement of Time class

lablnet:
Fix for https://github.com/zestframework/Zest_Framework/issues/234

### DIFF
--- a/src/Time/Time.php
+++ b/src/Time/Time.php
@@ -21,6 +21,63 @@ use Zest\Contracts\Time\Time as TimeContract;
 
 class Time implements TimeContract
 {
+
+    /**
+     * Date format.
+     *
+     * @since 3.0.0
+     *
+     * @var string
+    */
+    private $dateFormat;
+
+    /**
+     * Set the time format.
+     *
+     * @param mixe $format Valiud Date/Time format.
+     *
+     * @since 3.0.0
+     *
+     * @return object
+     */
+     public function setDateFormat($format)
+     {
+        if ($this->validateDateFormat($fornat)) {
+            $this->dateFormat = $format;
+
+            return $this;
+        }
+
+        throw new \InvalidArgumentException("The format {$format} is not valid format", 500);
+        
+     }
+
+    /**
+     * Get the time format.
+     *
+     * @since 3.0.0
+     *
+     * @return string
+     */
+     public function getDateFormat()
+     {
+        return $this->dateFormat;        
+     }
+
+    /**
+     * Validate the time format.
+     *
+     * @param mixe $format Valiud Date/Time format.
+     *
+     * @since 3.0.0
+     *
+     * @return bool
+     */
+     public function setDateFormat($format)
+     {
+        //code        
+     }
+
     /**
      * Get "now" time.
      *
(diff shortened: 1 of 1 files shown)


lablnet:
What about `validation` Method? @peter279k

src/Time/Time.php
```diff
@@ -21,6 +21,63 @@
...
+    /**
+     * Validate the time format.
+     *
+     * @param mixe $format Valiud Date/Time format.
+     *
+     * @since 3.0.0
+     *
+     * @return bool
+     */
+     public function setDateFormat($format)
```
peter279k: What kinds of types for `$format`? The `DateTime` class and `string`?

src/Time/Time.php
```diff
@@ -82,8 +139,10 @@ public static function timestampToGmt($time = null)
         $time = $time ?? time();
         $dateTime = new \DateTime();
         $dateTime->setTimestamp($time)->modify('+2 hours');
+        $format = isset($this->dateFormat) ? $format : 'd/m/Y H:i:s';
```
peter279k, line 146: We can consider letting the `$this->dateFormat` be the default date time format string.

Then this line should be unnecessary.

src/Time/Time.php
```diff
@@ -101,8 +160,9 @@ private static function timestampToDate($time = null)
         $time = self::formatTime($time);
         $dateTime = new \DateTime();
         $dateTime->setTimestamp($time);
+        $format = isset($this->dateFormat) ? $format : 'd-m-Y H:i:s';
```
peter279k, line 167: It has the same issue that I mention on the previous comment thread.

src/Time/Time.php
lablnet: So it mean it should has default value as well, or user should need to set it?

src/Time/Time.php
lablnet: Typo

lablnet: Yeah its Date Tine class

peter279k:
BTW, the PR title should be `Enhancement`, not `Entrenchment`.

src/Time/Time.php
peter279k: That's right.

src/Time/Time.php
peter279k: @Lablnet, thanks.

But it seems that the `$format` type is `mixed` inside the this method comment annotation.

The `mixed` type comment annotation means this `$format` can allow two or more types.

lablnet: yes i will fix this but how/what about the validation process

peter279k: Could you illustrate the `validation` approach?

lablnet: idk you mentioned in issue

src/Time/Time.php
```diff
@@ -21,6 +21,63 @@
...
+    /**
+     * Validate the time format.
+     *
+     * @param mixe $format Valiud Date/Time format.
+     *
+     * @since 3.0.0
+     *
+     * @return bool
+     */
+     public function validateDateFormat($format)
```
peter279k, line 76: Here is the [reference ](https://www.php.net/manual/en/function.checkdate.php#113205)about validating date format approach.

lablnet: oh, we need to validate Date with format or just format

is that what you propose

> Add the validateDateFormat method when setting setDateFormat

peter279k: I think we can validate date format, not just for the date format string.

Issue #247, opened 2019-09-28
Enchantment of str class


lablnet:
As title, enchantment of str Class

Need encoding support.

May need more methods like 
1. hasUpperCase
2. hasLowerCase
3. ConvertCase
4. etc...

ghost:
hey, 
do you have any coding guidelines? 
And do you have any other special wishes what kind of methods you need or should I just add everything that comes into my mind?

lablnet:
> hey,
> do you have any coding guidelines?
> And do you have any other special wishes what kind of methods you need or should I just add everything that comes into my mind?


Yes have coding guide
1.You never allow add third-party Composer dependencies or classes
2.You allow to improve existing features or add new features
3. You allow update Zest framework code or core code
4. When you want send pull request. You have provide brief description about code either you update exsting code or adding new feature.
In short every change you made should be standalone.

You may add everything that is in your mind
You may also consider this

hasUpperCase
hasLowerCase
ConvertCase

Thanks you for being interest in Contribution of Zest Framework

peter279k:
@Lablnet, we also consider adding the unit tests for `Str` class.

And it can be reliable for `Str` class method behaviors.

lablnet:
> @Lablnet, we also consider adding the unit tests for `Str` class.
> 
> And it can be reliable for `Str` class method behaviors.

Yeah, it already been their

look

https://github.com/zestframework/Zest_Framework/blob/master/Tests/Data/StrTest.php

PR #248, opened 2019-09-28
Adding Encoding support

lablnet:
https://github.com/zestframework/Zest_Framework/issues/247

### DIFF
--- a/src/Contracts/Data/Str.php
+++ b/src/Contracts/Data/Str.php
@@ -22,18 +22,19 @@ interface Str
      * Reverse the string.
      *
      * @param string $str String to be evaluated.
+     * @param string $encoding Valid encoding.
      *
      * @since 3.0.0
      *
      * @return string
      */
-    public static function reverse(string $str) :string;
+    public static function reverse(string $str, $encoding = null) :string;
 
     /**
      * Concat the strings.
      *
-     * @param string $g   With concat.
-     * @param string $str String to concat.
+     * @param string $g    With concat. 
+     * @param string $str  String to concat.
      *
      * @since 3.0.0
      *
@@ -45,10 +46,11 @@ interface Str
      * Count the string.
      *
      * @param string $str String to be counted.
+     * @param string $encoding Valid encoding.
      *
      * @since 3.0.0
      *
      * @return int
      */
-    public static function count(string $str);
+    public static function count(string $str, $encoding = null);
 }
--- a/src/Contracts/Data/Str.php
+++ b/src/Contracts/Data/Str.php
@@ -21,7 +21,7 @@ interface Str
     /**
      * Reverse the string.
      *
-     * @param string $str String to be evaluated.
+     * @param string $str      String to be evaluated.
      * @param string $encoding Valid encoding.
      *
      * @since 3.0.0
@@ -45,7 +45,7 @@ interface Str
     /**
      * Count the string.
      *
-     * @param string $str String to be counted.
+     * @param string $str      String to be counted.
      * @param string $encoding Valid encoding.
      *
      * @since 3.0.0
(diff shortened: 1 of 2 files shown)


src/Contracts/Data/Str.php
```diff
@@ -21,19 +21,20 @@ interface Str
...
-     * @param string $str String to be evaluated.
+     * @param string $str      String to be evaluated.
+     * @param string $encoding Valid encoding.
      *
      * @since 3.0.0
      *
      * @return string
      */
-    public static function reverse(string $str) :string;
+    public static function reverse(string $str, $encoding = null) :string;
```
peter279k, line 31: It looks like the `$encoding` type is `string`, and I think the `$encoding` default value should be empty.

If we set the `$encoding` is `null`,the `$encoding` `@param` annotation should not only `string`.

src/Contracts/Data/Str.php
```diff
@@ -44,11 +45,12 @@ public static function concat($g, ...$str);
...
-     * @param string $str String to be counted.
+     * @param string $str      String to be counted.
+     * @param string $encoding Valid encoding.
      *
      * @since 3.0.0
      *
      * @return int
      */
-    public static function count(string $str);
+    public static function count(string $str, $encoding = null);
```
peter279k, line 55: The situation is same as previous comment.

src/Data/Str.php
```diff
@@ -21,23 +21,39 @@
...
      *
-     * @param string $str String to be evaluated.
+     * @param string $encoding Valid encoding.
      *
      * @since 3.0.0
      *
      * @return string
      */
-    public static function reverse(string $str) :string
+    private static function encoding($encoding = null) :string
```
peter279k, line 32: This situation I mention on previous comment thread..

src/Data/Str.php
```diff
@@ -21,23 +21,39 @@
...
     {
-        if (function_exists('strrev')) {
-            return strrev($str);
-        }
+        return $encoding ?: \mb_internal_encoding();
```
peter279k, line 34: The `mb_internal_encoding` function is available on `php-mbstring` extension.

It should check the `mbstring` is available during composer installation.

src/Contracts/Data/Str.php
lablnet: SO it should be mixed?

src/Data/Str.php
```diff
@@ -61,16 +77,17 @@ public static function concat($g, ...$str)
...
-     * @param string $str String to be counted.
+     * @param string $str      String to be counted.
+     * @param string $encoding Valid encoding.
      *
      * @since 3.0.0
      *
      * @return int
      */
-    public static function count(string $str)
+    public static function count(string $str, $encoding = null)
```
peter279k, line 87: This situation I've mentioned on previous comment.

src/Data/Str.php
lablnet: If not available we have to install it?

I think you may send PR, in your available time

src/Contracts/Data/Str.php
peter279k: The `param` should be `$encoding|null`.

And function name should be `public static function reverse(string $str, ?$encoding) :string;`.

This syntax is available on `php-7.1+`.

peter279k: The reference is available on [PHPDoc](https://docs.phpdoc.org/guides/types.html#multiple-types-combined).

lablnet: Ok thanks you

so this syntax `?$encoding` tell it can null means to say optional right?

peter279k: Yes. That's right.

PR #258, opened 2019-10-01
hasUppercase and hasLowercase methods

peter279k:
# How to report a issue on GitHub
- It's related to issue #247.
- Add the `hasUppercase` and `hasLowercase` methods.
- They also include the tests.

### DIFF
--- a/src/Data/Str.php
+++ b/src/Data/Str.php
@@ -63,11 +63,11 @@ class Str implements StrContract
      * Concat the strings.
      *
      * @param string $g   With concat.
-     * @param string $str String to concat.
+     * @param array  $str String to concat.
      *
      * @since 3.0.0
      *
-     * @return bool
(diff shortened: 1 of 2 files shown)


src/Data/Str.php
```diff
@@ -63,11 +63,11 @@ public static function reverse(string $str, $encoding = null) :string
      * Concat the strings.
      *
      * @param string $g   With concat.
-     * @param string $str String to concat.
+     * @param array  $str String to concat.
      *
      * @since 3.0.0
      *
-     * @return bool
+     * @return string
```
peter279k, line 70: The `$str` param type is `array` because this method uses `...` syntax.

The sample code for `...` syntax is as follows:

image

peter279k:
Going ahead with PR #260.

Issue #243, opened 2019-09-28
Test's for container class


lablnet:
As title, need test's for container class

PR #263, opened 2019-10-01
Extends the use of hasUpperCase and hasLowerCase function, adds new test cases

hasukmistry:
- Extends use of hasUpperCase and hasLowerCase functions.
- Adds new test cases

PR #264, opened 2019-10-01
Test's for container class #243

hasukmistry:
It adds initial class file for writing test methods for the container.

### DIFF
--- a/Tests/Container/ContainerTest.php
+++ b/Tests/Container/ContainerTest.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Framework\Tests;
+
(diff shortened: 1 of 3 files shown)


lablnet (approved):

lablnet:
What's different between this PR and #263

Only difference I see in this you remove 'use' statement from tests.

I think this PR should be close due to conflict with previous PR

hasukmistry:
@Lablnet I think I created new branch from my forked master. That's why it shows that many changes. I just wanted to add a class file for testing container.

lablnet:
Hmm you may open new PR

PR #265, opened 2019-10-01
Adds class file for container test

hasukmistry:


lablnet:
`Use` Statement is required

`use Zest\Container\Container`

Tests/Container/ContainerTest.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Framework\Tests;
+
+use PHPUnit\Framework\TestCase;
+
+class ContainerTest extends TestCase
+{
+}
```
peter279k, line 24: This Container test doesn't have any tests.

Please add at least one test to do `Container` class test.

lablnet: Yes you're right

lablnet (changes requested):
Add some `tests` this is nothing you do.

Tests/Container/ContainerTest.php
hasukmistry: `use Zest\Container\Container` I added this statement first.  Because we did not have any test cases there, the usage of that class did not occur in it. So an automatic code check was complaining about it. Then later I removed that statement. I can add a test case to it. Can anyone suggest a test here? :) @peter279k @Lablnet

lablnet: You may looks to others frameworks test in case of `Laravel` ....

hasukmistry: @Lablnet  and @peter279k, I have implemented one test for container and also modified two functions in container. It will be great if you look at it. :)

hasukmistry:
I will create new PR for this.

PR #266, opened 2019-10-01
Enchantment of str class #247 - Adds convertStringToLowercase, convertStringToUppercase, getSubstring

k-kaufmann:
I wrapped the php the php fnctions for lower and uppercase and implemented my version of substring.

The substring method doesn't work with UTF-8 because I use array access to the string


lablnet:
This PR is conflicting with #267 , also you 

This PR is now closed, you may open new PR to add `getSubstring` method

Thanks

PR #267, opened 2019-10-01
Adds convert case method in str class and adds tests

hasukmistry:


lablnet:
@hasukmistry  Thanks you so much for your time

PR #269, opened 2019-10-01
Member ttl has private access

wesolowski:
Member has private access and can`t not by use in:
- APC@saveItem
- APCU@saveItem
- FileCache@saveItem
- Memcache@saveItem
- Memcached@saveItem
- Redis@saveItem
- SessionCache@saveItem


lablnet:
@wesolowski  thanks you so much.

PR #270, opened 2019-10-01
Bugfix: $key is private

wesolowski:
Member: $key is private and can`t not by use in SodiumEncryption


lablnet:
@wesolowski  Thanks you so much for your time to contribute

Really appropriated.

Issue #271, opened 2019-10-02
Enchantment of \Str class


lablnet:
As title need more method to make class more amazing.

We may consider something following methods

- `isBase64`
- `substring`
- `stripWhitespace`
- `repeat`
- `slice`
- `shuffel`

PR #272, opened 2019-10-02
Add isBase64 and substring methods to Str class

2mt-heuser:
As requested in https://github.com/zestframework/Zest_Framework/issues/271 I added two new statics to:

a) check if a string is a valid base64 encoded string
b) return a substring of given string

and added corresponding tests as well.

### DIFF
--- a/src/Data/Str.php
+++ b/src/Data/Str.php
@@ -169,4 +169,31 @@ class Str implements StrContract
         // returns original string when mbstring is not active.
         return $str;
     }
+
+    /**
+     * Checks if given string is a valid Base64 encoded string
+     *
+     * @param string $str String to be tested
+     *
+     * @return bool wether it is a valid base64 encoded string or not
+     */
+    public static function isBase64(string $str): bool
+    {
+        return ($str === base64_encode(base64_decode($str)));
+    }
+
+    /**
+     * Return only a portion of given string
+     *
+     * @param string $str input string to process
+     * @param int $start where to start the cut
+     * @param int|null $length how many characters to return
+     * @param string $encoding optional encoding to use
+     *
+     * @return string
+     */
+    public static function substring(string $str, int $start, int $length = null, string $encoding = 'UTF-8'): string
+    {
+        return mb_substr($str, $start, $length, $encoding);
+    }
 }
--- a/src/Data/Str.php
+++ b/src/Data/Str.php
@@ -185,10 +185,10 @@ class Str implements StrContract
     /**
      * Return only a portion of given string.
      *
-     * @param string   $str      input string to process
-     * @param int      $start    where to start the cut
-     * @param int|null $length   how many characters to return
-     * @param string   $encoding optional encoding to us
+     * @param string      $str      input string to process
+     * @param int         $start    where to start the cut
+     * @param int|null    $length   how many characters to return
+     * @param string|null $encoding optional encoding to use
      *
      * @return string
      */
--- a/src/Data/Str.php
+++ b/src/Data/Str.php
@@ -171,7 +171,7 @@ class Str implements StrContract
     }
 
     /**
-     * Checks if given string is a valid Base64 encoded string
+     * Checks if given string is a valid Base64 encoded string.
      *
      * @param string $str String to be tested
      *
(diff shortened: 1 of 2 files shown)


src/Data/Str.php
```diff
@@ -169,4 +169,31 @@ public static function ConvertCase(string $str, $encoding = null)
...
+     * @param string $str input string to process
+     * @param int $start where to start the cut
+     * @param int|null $length how many characters to return
+     * @param string $encoding optional encoding to use
+     *
+     * @return string
+     */
+    public static function substring(string $str, int $start, int $length = null, string $encoding = 'UTF-8'): string
+    {
+        return mb_substr($str, $start, $length, $encoding);
```
lablnet: the last argument should be like `self::encoding($encoding);`

The value of encoding argument in function should be null or `?$encoding`

2mt-heuser: thanks for the hint, totally overlooked how it was done in the other functions... 

should be fixed now

peter279k:
The Travis CI build is failed, but it's not related to this PR.

I will ignore CI build at this moment and focus on reviewing code changes.

src/Data/Str.php
```diff
@@ -169,4 +169,31 @@ public static function ConvertCase(string $str, $encoding = null)
...
+     * Return only a portion of given string.
+     *
+     * @param string   $str      input string to process
+     * @param int      $start    where to start the cut
+     * @param int|null $length   how many characters to return
+     * @param string   $encoding optional encoding to us
+     *
+     * @return string
+     */
+    public static function substring(string $str, int $start, int $length = null, $encoding = null): string
```
peter279k, line 195: IMHO, the `substring` method should be changed as follows:

```php
public static function substring(string $str, int $start, ?int $length, ?$encoding): string
```

That means the `$length` and `$encoding` will be possible to be `null`.

And this syntax is available since `php-7.1`.

src/Data/Str.php
```diff
@@ -169,4 +169,31 @@ public static function ConvertCase(string $str, $encoding = null)
...
+        return $str === base64_encode(base64_decode($str));
+    }
+
+    /**
+     * Return only a portion of given string.
+     *
+     * @param string   $str      input string to process
+     * @param int      $start    where to start the cut
+     * @param int|null $length   how many characters to return
+     * @param string   $encoding optional encoding to us
```
peter279k: This `@param` comment annotation should change into `@param string|null   $encoding optional encoding to us`.

peter279k (changes requested):
Some request changes should be concerned.

They're related to comment annotation and type hint approach.

Thanks for your contributions again :).

src/Data/Str.php
2mt-heuser: with this change one would need to explicitly set encoding to null instead of omitting it, else:

ArgumentCountError: Too few arguments to function Zest\Data\Str::substring(), 3 passed in /c/workspace/hacktoberfest/Zest_Framework/Tests/Data/StrTest.php on line 62 and exactly 4 expected

peter279k: Hi @2mt-heuser, sorry. My suggested code is incorrect. Please try this:

```php
public static function substring(string $str, int $start, ?int $length, ?string $encoding): string
```

2mt-heuser: @peter279k already did. but the question mark only allows null or string aka making the parameter required instead of optional.
Without a default value it will crash if you just call the function with 3 parameters.

```php
public static function substring(string $str, int $start, ?int $length, ?string $encoding = null)
```
this would work, but in thise case ?string wouldn't make any sense as we're just setting it to null anyway, so we could just leave it without the ?
```php
public static function substring(string $str, int $start, ?int $length, string $encoding = null)
```

Same thing applies to ?int $length. If we only call substring with 2 parameters it should use default encoding and return the string from $start until the end but it will crash with ?int $length and not setting it to null
```php
$this->assertSame('def', Str::substring('abcdef', 3));
```
So best bet would be to stick to the current version, right?!
Sounds okay to you?

src/Data/Str.php
2mt-heuser: will change it and fix my typo at the end as well. should be "to use"

lablnet: yes

lablnet: Let's fix so i will merge this PR

src/Data/Str.php
peter279k: Going ahead with this.

src/Data/Str.php
2mt-heuser: Done and also added another testcase for the above discussion about optional parameters

peter279k (approved):
I think it's good and it's time to be merged.

lablnet (approved):

lablnet:
@2mt-heuser  Thanks you so much for your time to contribute

Really appropriated.

2mt-heuser:
> @2mt-heuser Thanks you so much for your time to contribute
> 
> Really appropriated.

You're welcome. 
When I find more time between work I'll take a look at some other methods and/or issues.

Issue #274, opened 2019-10-02
Wrong Behaviours of Str \hasUpperCase and \hasLowerCase


lablnet:
Consider the following string

`#-<ae|9_+0a*a!Q?@(z,3R>.BAb)^%R&*+~RG:$`

### Upper Case `Regx`

![image](https://user-images.githubusercontent.com/27757785/66021094-df444780-e502-11e9-9066-30f88837fc8e.png)

same regx use in LowerCase

peter279k:
I'm curious about the `hasLowercase` and `hashUppercase` method implementations.

Should we consider only about the English string lowercase/uppercase conversion?

If possible, we can change into following regular expression:

`hasUppercase`:

```
/([A-Z])/
```

`hasLowercase`:

```
/([a-z])/
```

![image](https://user-images.githubusercontent.com/9021747/66023105-baf86300-e522-11e9-976e-f390cc2f667a.png)

lablnet:
> I'm curious about the `hasLowercase` and `hashUppercase` method implementations.
> 
> Should we consider only about the English string lowercase/uppercase conversion?
> 
> If possible, we can change into following regular expression:
> 
> `hasUppercase`:
> 
> ```
> /([A-Z])/
> ```
> 
> `hasLowercase`:
> 
> ```
> /([a-z])/
> ```
> 
> ![image](https://user-images.githubusercontent.com/9021747/66023105-baf86300-e522-11e9-976e-f390cc2f667a.png)

We need to keep in mind encoding as well, in that case UniCode

PR #275, opened 2019-10-02
Fix for #274

lablnet:
https://github.com/zestframework/Zest_Framework/issues/274

### DIFF
--- a/src/Data/Str.php
+++ b/src/Data/Str.php
@@ -112,7 +112,7 @@ class Str implements StrContract
         $all_words = preg_split('/\s+/', $str);
 
         foreach ($all_words as $word) {
-            if (preg_match_all('~^\p{Lu}~u', $word)) {
+            if (preg_match_all('~\p{Lu}~u', $word)) {
                 // There is one upper
                 return true;
             }
@@ -133,8 +133,8 @@ class Str implements StrContract
         $all_words = preg_split('/\s+/', $str);
 
         foreach ($all_words as $word) {
-            if (!preg_match_all('~^\p{Lu}~u', $word)) {
-                // There is one upper
+            if (!preg_match_all('~\p{Ll}~u', $word)) {
+                // There is one lower
                 return true;
             }
         }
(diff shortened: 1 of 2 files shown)


lablnet:
@peter279k  I do not know why test failing.

peter279k:
The Travis CI build is failed because the assertions are failed.

It seems that the regular expression patterns are not correct. And it should be checked.

The failed assertion messages are as follows:

```
1) Framework\Tests\StrTest::testHasLowerCase
/home/travis/build/zestframework/Zest_Framework/Tests/Data/StrTest.php:39
```

lablnet:
> The Travis CI build is failed because the assertions are failed.
> 
> It seems that the regular expression patterns are not correct. And it should be checked.
> 
> The failed assertion messages are as follows:
> 
> ```
> 1) Framework\Tests\StrTest::testHasLowerCase
> Failed asserting that false is true.
> /home/travis/build/zestframework/Zest_Framework/Tests/Data/StrTest.php:39
> ```

Am sure "Regx" is correct, you may try to use online regx editor and manually check that in your Home.php" file it produce right result

i do not know what the hack problem with UnitTest

src/Data/Str.php
```diff
@@ -133,8 +133,8 @@ public static function hasLowerCase(string $str)
         $all_words = preg_split('/\s+/', $str);
 
         foreach ($all_words as $word) {
-            if (!preg_match_all('~^\p{Lu}~u', $word)) {
-                // There is one upper
+            if (!preg_match_all('~\p{Ll}~u', $word)) {
```
peter279k, line 136: The problem is about the `preg_match_all` returns value.

The `preg_match_all` will return integer, I think we don't use the `!preg_match_all` to check whether the return value is.

It should use `preg_match_all(.....) === 0` condition.

Reference: https://www.php.net/manual/en/function.preg-match-all.php

lablnet:
This `PR` is now closed due to better one #280

PR #277, opened 2019-10-02
Update StrTest.php

lablnet:


Tests/Data/StrTest.php
```diff
@@ -35,7 +35,7 @@ public function testHasUpperCase()
 
     public function testHasLowerCase()
     {
-        $this->assertFalse(Str::hasLowerCase('CAMEL'));
+        //$this->assertFalse(Str::hasLowerCase('CAMEL'));
```
peter279k, line 38: Why adding comment for this assertion?

PR #278, opened 2019-10-02
Add String methods

YWatchman:
As issue #271 stated, I've added all the static functions.


lablnet:
This `PR` is conflicts 
And the implementation should be concern encoding too

This `PR` is now closed

Thanks for your time to contribute, fix the above mentioned error and open new `PR`

PR #279, opened 2019-10-02
Add String methods 

YWatchman:
As issue #271 stated, I've added all the static functions.

Merged with upstream branch for multibyte encoding, please provide feedback don't close it yet.

### DIFF
--- a/Tests/Data/StrTest.php
+++ b/Tests/Data/StrTest.php
@@ -92,4 +92,9 @@ class StrTest extends TestCase
         );
         $this->assertNotSame('good shit', Str::slice('This is good shit', 8, 2));
     }
+
+    public function testShuffle()
+    {
+        $this->assertNotSame('This is string is not shuffled', Str::shuffle('This string is not shuffled'));
+    }
 }
--- a/Tests/Data/StrTest.php
+++ b/Tests/Data/StrTest.php
@@ -48,4 +48,10 @@ class StrTest extends TestCase
         $this->assertSame('UPPERcase', Str::ConvertCase('upperCASE', 'UTF-8'));
         $this->assertSame('LOWERCASE IN string', Str::ConvertCase('lowercase in STRING', 'UTF-8'));
     }
+
+    public function testisBase64()
+    {
+        $this->assertFalse(Str::isBase64('4rdHFh%2BHYoS8oLdVvbUzEVqB8Lvm7kSPnuwF0AAABYQ%3D'));
+        $this->assertTrue(Str::isBase64('YW1vdXJmYXlh'));
+    }
 }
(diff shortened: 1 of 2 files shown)


Tests/Data/StrTest.php
```diff
@@ -49,8 +49,10 @@ public function testConvertCase()
         $this->assertSame('LOWERCASE IN string', Str::ConvertCase('lowercase in STRING', 'UTF-8'));
     }
 
-    public function testIsBase64()
+    public function testisBase64()
```
peter279k: This changes I think should be reverted because of camel case naming approach.

Tests/Data/StrTest.php
```diff
@@ -59,11 +61,50 @@ public function testIsBase64()
...
+    public function testStripWhitespaces()
+    {
+        $this->assertSame('This string is stripped', Str::stripWhitespaces(' This string is stripped '));
+        $this->assertNotSame(' This string is stripped ', Str::stripWhitespaces(' This string is stripped '));
+    }
+
+    public function testRepeat()
+    {
+        $this->assertSame('--', Str::repeat('--', 1));
+        $this->assertNotSame('--', Str::repeat('--', 3));
```
peter279k, line 84: Could you add another `assertSame` assertion about `Str::repeat('--', 3)`?

src/Data/Str.php
```diff
@@ -171,29 +171,107 @@ public static function ConvertCase(string $str, $encoding = null)
...
+    public static function isBase64(string $string)
+    {
+        // Check if there are valid base64 characters
+        if (!preg_match('/^[a-zA-Z0-9\/\r\n+]*={0,2}$/', $string)) {
+            return false;
+        }
+
+        // Decode the string in strict mode and check the results
+        $decoded = base64_decode($string, true);
+        if (false === $decoded) {
```
peter279k: @Lablnet, I think we don't use the YODA condition.

peter279k (changes requested):
Some request changes should be concerned.

Tests/Data/StrTest.php
YWatchman: Merge fail, will do.

YWatchman:
Please review @peter279k

src/Data/Str.php
```diff
@@ -171,29 +171,107 @@ public static function ConvertCase(string $str, $encoding = null)
...
-     * @param string $str String to be tested
+     * @param string $string String to be tested
      *
-     * @return bool wether it is a valid base64 encoded string or not
+     * @return bool
+     */
+    public static function isBase64(string $string)
+    {
+        // Check if there are valid base64 characters
+        if (!preg_match('/^[a-zA-Z0-9\/\r\n+]*={0,2}$/', $string)) {
```
peter279k: As far as I know, the `preg_match` function will return `integer` value.

I think it can consider using the `preg_match('/^[a-zA-Z0-9\/\r\n+]*={0,2}$/', $string) === 0` condition instead.

YWatchman: preg_match() returns either 0 or a bool. 0 and 1 are both considered boolean, I'll just follow your recommendation and do make a strict conditional.

peter279k (approved):
It looks good to me.

@Lablnet, could you review this PR?

lablnet (approved):

lablnet:
@YWatchman  Thanks you so much for your time to contribute .

Really appropriated.

PR #280, opened 2019-10-02
Fix behaviour of hasUppercase and hasLowercase

2mt-heuser:
Regarding https://github.com/zestframework/Zest_Framework/issues/274 how about this implementation of hasUpper/Lower?

Would that have any drawbacks?

### DIFF
--- a/src/Data/Str.php
+++ b/src/Data/Str.php
@@ -118,6 +118,7 @@ class Str implements StrContract
      *
      * @param string      $str      String to be checked.
      * @param string|null $encoding Valid encoding.
+     * 
      * @return bool
      */
     public static function hasLowerCase(string $str, string $encoding = null): bool
--- a/src/Data/Str.php
+++ b/src/Data/Str.php
@@ -103,43 +103,26 @@ class Str implements StrContract
     /**
      * Check if string has atleast one uppercase.
      *
-     * @param string $str String to be checked.
+     * @param string      $str      String to be checked.
+     * @param string|null $encoding Valid encoding.
      *
      * @return bool
      */
-    public static function hasUpperCase(string $str)
+    public static function hasUpperCase(string $str, string $encoding = null): bool
     {
-        $all_words = preg_split('/\s+/', $str);
-
-        foreach ($all_words as $word) {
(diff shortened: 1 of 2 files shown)


lablnet (approved):

src/Data/Str.php
```diff
@@ -103,43 +103,27 @@ public static function count(string $str, $encoding = null)
...
      * Check if string has atleast one uppercase.
      *
-     * @param string $str String to be checked.
+     * @param string      $str      String to be checked.
+     * @param string|null $encoding Valid encoding.
      *
      * @return bool
      */
-    public static function hasUpperCase(string $str)
+    public static function hasUpperCase(string $str, string $encoding = null): bool
```
peter279k, line 111: How about using following code?

```php
public static function hasUpperCase(string $str, ?string $encoding = null): bool
```

2mt-heuser: would have the exact same call-signature / implication?

when providing the default ```$encoding = null``` we don't need the ? to mark is as string|null

but I could change it to be in line with count() etc where there's no typehint string and only the default value if you'd prefer that.
```php
public static function reverse(string $str, $encoding = null) :string
public static function count(string $str, $encoding = null)
```

or add typehints to all the functions so they're inline with each other

peter279k: Hi, thanks for your explanation.

I don't prefer any of this type hints and @Lablnet may decide to use.

I care about that is about consistency.

Once we decide to use one of type hint style. It should be used on whole repository.

lablnet: I think we can continue as we are going with default type hinting..

2mt-heuser: With 7.4 and typed properties around the corner I kinda like transitioning to everything typed but nontheless, your choice. If you want them gone, it's just one commit from me away :-)

lablnet: No thanks for your time am going to Merger this PR, but lets first approve by @peter279k

peter279k (approved):

lablnet:
Thanks you so much for your time .

PR #281, opened 2019-10-02
feat: add the option to output assoc arrays fixes #261

### LINKED ISSUE #261: Bug in \Arrays class
Let consider the following `DataSet`

```php
        $dataSets =    [
       'files' => [
           //Default file mine type
           'mine' => [
               'type' => [
                   'application/x-zip-compressed',
                   'application/msword',
                   'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
                   'image/gif',
                   'image/jpeg',
                   'image/jpeg',
                   'audio/mpeg',
                   'video/mp4',
                   'application/pdf',
                   'image/png',
                   'application/zip',
                   'application/et-stream',
                   'image/x-icon',
                   'image/icon',
                   'image/svg+xml',                
               ],
           ],
   
           //Default types
           'types' => [
               'image' => ['jpg', 'png', 'jpeg', 'gif', 'ico', 'svg'],
               'zip'   => ['zip', 'tar', '7zip', 'rar'],
               'docs'  => ['pdf', 'docs', 'docx'],
               'media' => ['mp4', 'mp3', 'wav', '3gp'],
           ],
        ]];
```

then consider following statement

```php
        $array = new Arrays;
        var_dump($array->dot($dataSets));
```

It produce something like that result
image

Which is wrong, it should produce something like that

'files.mine.type. => array.'

There should be flag if we want to return array from specific key.

@peter279k  what do you think?

https://github.com/zestframework/Zest_Framework/blob/master/src/Data/Arrays.php#L269

ReeceM:
This PR will add the ability to the `Zest\Data\Arrays::class` to output associative arrays when they are present at the end of a file.

For references and further discussion of this please look at #261 for the test data set.

The output of using this edited method is backwards compatible, it can be opted into for getting the different output.

An example: 
The syntax to access that would be:
```php
$array = new Arrays;
var_export($array->dot($dataSets, true));
```

```php
[
     "files.mine.type" => [
       "application/x-zip-compressed",
       "application/msword",
       "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
       "image/gif",
       "image/jpeg",
       "image/jpeg",
       "audio/mpeg",
       "video/mp4",
       "application/pdf",
       "image/png",
       "application/zip",
       "application/et-stream",
       "image/x-icon",
       "image/icon",
       "image/svg+xml",
     ],
     "files.types.image" => [
       "jpg",
       "png",
       "jpeg",
       "gif",
       "ico",
       "svg",
     ],
     "files.types.zip" => [
       "zip",
       "tar",
       "7zip",
       "rar",
     ],
     "files.types.docs" => [
       "pdf",
       "docs",
       "docx",
     ],
     "files.types.media" => [
       "mp4",
       "mp3",
       "wav",
       "3gp",
     ],
   ]
```

### DIFF
--- a/src/Data/Arrays.php
+++ b/src/Data/Arrays.php
@@ -274,10 +274,10 @@ class Arrays implements ArraysContract
     /**
      * Converted a multi-dimensional associative array with `operator`.
      *
-     * @param array  $arrays Arrays.
-     * @param string $opr    Operator.
-     * @param bool  $assocOutput Switch to output assoc arrays.
-     * @param string $_key the previous key of the object
+     * @param array  $arrays      Arrays.
+     * @param string $opr         Operator.
+     * @param bool   $assocOutput Switch to output assoc arrays.
+     * @param string $_key        the previous key of the object
      *
      * @since 3.0.0
      *
@@ -287,7 +287,7 @@ class Arrays implements ArraysContract
     {
         $results = [];
         foreach ($arrays as $key => $value) {
-            $key = ($assocOutput === true ? $_key . $key : $key);
+            $key = ($assocOutput === true ? $_key.$key : $key);
             if (self::isReallyArray($value) === true) {
                 $assocKey = ($assocOutput === true ? $opr : $key.$opr);
                 $results = array_merge(
@@ -296,7 +296,7 @@ class Arrays implements ArraysContract
                         $value,
                         $assocKey,
                         $assocOutput,
-                        $key . $opr
+                        $key.$opr
                     )
                 );
             } elseif ($assocOutput === true) {
--- a/src/Data/Arrays.php
+++ b/src/Data/Arrays.php
@@ -301,12 +301,13 @@ class Arrays implements ArraysContract
                     )
                 );
             } elseif ($assocOutput === true) {
-                // gets pull off the last dot that is added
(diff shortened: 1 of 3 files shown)


ReeceM:
Adding fixes, don't merge

src/Data/Arrays.php
```diff
@@ -260,33 +260,49 @@ public static function multiToAssoc(array $arrays)
...
+                    self::multiToAssocWithSpecificOpr(
+                        $value,
+                        $assocKey,
+                        $assocOutput,
+                        $key.$opr
+                    )
+                );
+            } elseif ($assocOutput === true) {
+                // gets pull off the last dot that is added
+                $_key = preg_replace('/(\.$)/', '', $_key);
```
lablnet: It should not be dot specific.

lablnet (changes requested):

src/Data/Arrays.php
lablnet: You may consider Opr variable

ReeceM: Hi, thank you for that, I will change that now.

lablnet (approved):

lablnet:
@ReeceM  Thanks you so much for your time to contribute

Really appreciated.

PR #284, opened 2019-10-03
Create PasswordManipulationTest.php

lablnet:


Tests/Common/PasswordManipulationTest.php
```diff
@@ -0,0 +1,41 @@
...
+
+use PHPUnit\Framework\TestCase;
+use Zest\Common\PasswordManipulation;
+
+class PasswordManipulationTest extends TestCase
+{
+    public function generatePasswordTest()
+    {
+        $pass = new PasswordManipulation();
+        $this->assertNotEmpty($pass->generatePassword(11));
```
peter279k: It should add more assertions here.

For example, assert the password length and password pattern validation.

peter279k: It should add `test-based` prefix to name method name. And PHPUnit will know this methods are about tests.

For example, naming the `testGeneratePassword`.

peter279k (changes requested):
Some request changes that need to be concerned.

Tests/Common/PasswordManipulationTest.php
lablnet: > password

Password pattern is not need because isValid method do that already

Tests/Common/PasswordManipulationTest.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        $pass = new PasswordManipulation();
+        $pass->setLength(21);
+        $this->assertNotEmpty($this->getLength());
+    }
+
+    public function testGetLength()
+    {
+        $pass = new PasswordManipulation();
+        $this->assertNotEmpty($this->getLength());
```
peter279k: It should be `$pass->getLength`.

The Travis CI build is failed. They're as follows:

```
1) Framework\Tests\PasswordManipulationTest::testSetLength

Error: Call to undefined method Framework\Tests\PasswordManipulationTest::getLength()

/home/travis/build/zestframework/Zest_Framework/Tests/Common/PasswordManipulationTest.php:33

2) Framework\Tests\PasswordManipulationTest::testGetLength

Error: Call to undefined method Framework\Tests\PasswordManipulationTest::getLength()
```

Tests/Common/PasswordManipulationTest.php
```diff
@@ -0,0 +1,41 @@
...
+    public function testCountTes()
+    {
+        $pass = new PasswordManipulation();
+        $this->assertSame(6, $pass->len(123456));
+    }
+
+    public function testIsValid()
+    {
+        $pass = new PasswordManipulation();
+        $this->assertTrue($pass->isValid($pass->generatePassword(15)));
```
peter279k: It looks like the random password doesn't always have correct validation.

The failed assertion may happen:

```
1) Framework\Tests\PasswordManipulationTest::testIsValid

/home/travis/build/zestframework/Zest_Framework/Tests/Common/PasswordManipulationTest.php:25
```

peter279k: I have to say that the `isValid` method is problematic.

It should be concerned and enhanced.

lablnet: Lets consider this code

```php
       $password = new PasswordManipulation();
       $pass = $password->generatePassword();
       var_dump($pass);
       var_dump($password->isValid($pass));
```

For now forget unit Test, run this with Zest framework it will produce right result, what the probem with Unit test, idk

peter279k: No. That's not correct.

The password is generated randomly.

That is, the unit test will be failed probably.

Nobody can guarantee that the unit test will be failed.

Don't forget that the unit tests can help us to "find errors".

And it does, it helps us to find `isValid` method approach has been problematic.

lablnet: I think `generatePassword` is problematic that cant generate strong password , right?

peter279k: If `isValid` can help us to validate random password correctly, the `generatePassword` will be problematic probably.

lablnet: @peter279k  see all tests passed now, it mean problem with `generatePassword()` method

src/Common/PasswordManipulation.php
```diff
@@ -35,7 +35,7 @@ class PasswordManipulation
      *
      * @since 3.0.0
      *
-     * @return int
+     * @return mixed
```
peter279k: Why password length type is `integer`?

Could you explain that?

src/Common/PasswordManipulation.php
```diff
@@ -59,16 +59,11 @@ public function getLength()
      *
      * @since 2.9.7
      *
-     * @return mix-data
+     * @return mixed
```
peter279k: `generatePassword` will not only return `string` type?

src/Common/PasswordManipulation.php
lablnet: its return value, it return mixed, int or false

src/Common/PasswordManipulation.php
lablnet: it technically string, but with combinatin of specialChars, numbers etc

lablnet: is am right?

src/Common/PasswordManipulation.php
```diff
@@ -28,18 +28,22 @@ class PasswordManipulation
...
 
     /**
      * Set the password default length.
      *
+     * @param (int) $length Length of password.
+     *
      * @since 3.0.0
      *
-     * @return int
+     * @return object
```
peter279k: It should be `@return self` because it returns `$this` object.

src/Site/Site.php
```diff
@@ -196,17 +196,26 @@ public static function setLastSegment($segments)
...
-     * @param (int) $length Length of salts.
+     * @param (int)  $length  Length of salts.
+     * @param (bool) $special Should special chars include or not.
      *
      * @since 1.0.0
      *
      * @return string
      */
-    public static function salts($length)
+    public static function salts(int $length, $special = false)
```
peter279k: It should be `public static function salts(int $length, bool $special = false)`.

peter279k (approved):

Issue #293, opened 2019-10-04
After adding \Site\Key classs tests failing


lablnet:
The error log say that `Zest\Contracts\Site\Key` is not found

but this file in place and valid.

Also there is no such error in my `local`

Any idea why this happening?

hasukmistry:
Hi @Lablnet can you try using \ in front of it.  for example `use \Zest\Contracts\Site\Key` and also you have defined an interface with Key not key.

![Screenshot 2019-10-04 at 12 47 53 PM](https://user-images.githubusercontent.com/10345184/66202172-5580af80-e6a5-11e9-9a56-049bef4a2da3.png)

lablnet:
> Hi @Lablnet can you try using \ in front of it. for example `use \Zest\Contracts\Site\Key` and also you have defined an interface with Key not key.
> 
> ![Screenshot 2019-10-04 at 12 47 53 PM](https://user-images.githubusercontent.com/10345184/66202172-5580af80-e6a5-11e9-9a56-049bef4a2da3.png)

Am wondering why to put `\` because you may see other classes, there is no such issue with them, in case you may consider `Str`, `Arrays` etc...

Let's wait reply from @peter279k

lablnet:
This is still failing but in my local Unit Testing its passed

![image](https://user-images.githubusercontent.com/27757785/66204086-76f09400-e6c3-11e9-843e-8c27199ac6f6.png)


I do not know what to do!!!!!!!!!!!!!

hasukmistry:
@Lablnet Did you try to change `use Zest\Contracts\Site\key as KeyContract;` to `use Zest\Contracts\Site\Key as KeyContract;` in src\Site\Key.php

hasukmistry:
> This is still failing but in my local Unit Testing its passed
> 
> ![image](https://user-images.githubusercontent.com/27757785/66204086-76f09400-e6c3-11e9-843e-8c27199ac6f6.png)
> 
> I do not know what to do!!!!!!!!!!!!!

In your unit test, there is a correct import for `Key`

lablnet:
@

> > This is still failing but in my local Unit Testing its passed
> > ![image](https://user-images.githubusercontent.com/27757785/66204086-76f09400-e6c3-11e9-843e-8c27199ac6f6.png)
> > I do not know what to do!!!!!!!!!!!!!
> 
> In your unit test, there is a correct import for `Key`

Localhost is loosely environment

PR #296, opened 2019-10-04
Changes the interface statment to use

hasukmistry:
Bug: #293


lablnet:
@hasukmistry  Thanks you so much for your time to solve this issue

PR #298, opened 2019-10-05
Update FileInfo.php

ossamamehmood:
The `sha256` is more secured, then `md5`

### DIFF
--- a/src/Files/FileInfo.php
+++ b/src/Files/FileInfo.php
@@ -72,7 +72,7 @@ class FileInfo extends SplFileInfo
      *
      * @return string
      */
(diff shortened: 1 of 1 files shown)


lablnet (approved):

peter279k:
The `md5` and `sha-1` are not secure anymore.

Please deprecate them or throw exception to notify developers don't do that...

lablnet:
> The `md5` and `sha-1` are not secure anymore.
> 
> Please deprecate them or throw exception to notify developers don't do that...

That's right so which algo should be used to make it more secured?

you may open PR

PR #299, opened 2019-10-05
Added RSA key generation supports

lablnet:
Added support of Generate RSA keypair

@peter279k  what do you think?

### DIFF
--- a/Tests/Site/KeyTest.php
+++ b/Tests/Site/KeyTest.php
@@ -27,4 +27,10 @@ class KeyTest extends TestCase
         $key = Key::generateEncode(16);
         $this->assertEquals(32, mb_strlen($key, '8bit'));
     }
+
+    public function testGenerateRsa()
+    {
+    	$rsa = Key::generateRSAKey();
+    	$this->assertNotEmpty($rsa);
+    }
 }
--- a/src/Site/Key.php
+++ b/src/Site/Key.php
@@ -74,5 +74,27 @@ class Key implements keyContract
     public static function generateEncode(int $length = 32): string
     {
         return static::encode(self::generate($length));
-    }
+	}
+
+	/**
+	 * Generate RSA public and private keypair.
+	 *
+     * @since 3.0.0
+     *
+	 * @return array
+	 */
+	public static function generateRSAKey(): array
+	{
+		$rsa = openssl_pkey_new(array(
+			"private_key_bits" => 2048,
+			"private_key_type" => \OPENSSL_KEYTYPE_RSA,
+		));
+		openssl_pkey_export($rsa, $rsaPrivateKey);
+		$details = openssl_pkey_get_details($rsa);
+
+		return [
+			'public'  => $details['key'],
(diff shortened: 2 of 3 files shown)


Tests/Site/KeyTest.php
```diff
@@ -27,4 +27,10 @@ public function testGenerateEncode(): void
         $key = Key::generateEncode(16);
         $this->assertEquals(32, mb_strlen($key, '8bit'));
     }
+
+    public function testGenerateRsa()
+    {
+    	$rsa = Key::generateRSAKey();
+    	$this->assertNotEmpty($rsa);
```
peter279k, line 34: To make private/public key generating test meaningful, it should assert the public/private key is valid or verified.

src/Site/Key.php
```diff
@@ -74,5 +74,27 @@ public static function generate(int $length = 32): string
...
+	 * @return array
+	 */
+	public static function generateRSAKey(): array
+	{
+		$rsa = openssl_pkey_new(array(
+			"private_key_bits" => 2048,
+			"private_key_type" => \OPENSSL_KEYTYPE_RSA,
+		));
+		openssl_pkey_export($rsa, $rsaPrivateKey);
+		$details = openssl_pkey_get_details($rsa);
```
peter279k, line 93: Please check this [post](https://paragonie.com/blog/2018/04/protecting-rsa-based-protocols-against-adaptive-chosen-ciphertext-attacks#rsa-doing-it-right) to concern how to generate RSA key correctly.

peter279k (changes requested):
Some request changes are concerned:

- Coding style fix.
- RSA key generating test enhancement.
- RSA key generating approach enhancement.

peter279k:
Consider this [post](https://paragonie.com/blog/2016/12/everything-you-know-about-public-key-encryption-in-php-is-wrong), and I think we don't only use `RSA`.

It should use Hybrid `RSA`+`AES` encryption or other encryption approaches.

lablnet:
> AES

I thought we can use two approaches here

first generate `RSA` keypair 
Other is generate `RSA+AES` Hybrid combinations, so it will be upon user choose which to use....

what do you think?

Tests/Site/KeyTest.php
lablnet: It's not easy to validate key , am sure phpUnit has now such function to do that

peter279k: It's possible to validate key inside PHPUnit tests.

peter279k:
IMHO, using the hybrid `RSA + AES` encryption instead.

And don't use only `RSA` encryption approach.

Tests/Site/KeyTest.php
peter279k: If possible, it's fine to do this tests on this PR.

The meaningful tests can be on another PR.

lablnet:
This `PR` is now closed, will create new `PR` regarding this.

PR #309, opened 2019-10-18
Rename Expection to Exception.

Maikuolan:
"Expection" isn't a word in English. I'm assuming that that was a typo, and that "Exception" is what was meant. This pull request just renames the class and changes references from the old class name to the new class name accordingly. :-)


lablnet:
@Maikuolan  thanks you so much for your time to contribute

PR #311, opened 2020-08-25
fix for #220

### LINKED ISSUE #220: Archive component
**Is your feature request related to a problem? Please describe.**

According to the `Zest_Framework/src/Zip/`, it looks like the Archive only supports the `Zip`.

But it has other archive approaches. For example, the `bzip`, `gzip` and so on.

**Describe the solution you'd like**

Consider creating missed archive components.

Or try considering creating a `Archive` component to support all available archives in this framework.

lablnet:
https://github.com/zestframework/Zest_Framework/issues/220


peter279k (changes requested):
These classes seems to be done and my recommendations are as follows:

- Each archive classes should have their unit tests and it can make their behaviors verified and reliable.
- The StyleCI has been failed to check on the latest commit and it should be fixed.

lablnet:
@Maikuolan tests pass in `gh actions`

Maikuolan (approved):
LGTM. :-)

peter279k:
coding style [error ](https://github.styleci.io/analyses/yv0QnN?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)should be concerned.

src/Archive/Archive.php
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Set the adapter.
+     *
+     * @param (string) $adapter
+     *
+     * @since 3.0.0
+     *
+     * @return object
+     */
+    public function setAdapter($adapter)
```
peter279k: To be consistency, it should add the native type hint:

```PHP
public function setAdapter($adapter): self
```

lablnet:
> coding style [error ](https://github.styleci.io/analyses/yv0QnN?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)should be concerned.

there is no such error, its giving that in other file which is out of scope in sense of current PR

peter279k:
> > coding style [error ](https://github.styleci.io/analyses/yv0QnN?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)should be concerned.
> 
> there is no such error, its giving that in other file which is out of scope in sense of current PR

Right. Without coding style error, I think it should be good to merge this PR.

PR #321, opened 2020-11-24
Add an adapter interface.

Maikuolan:
When the intention exists to write numerous classes with the same or very similar methods, PHP interfaces are a useful means to ensure that accidents don't occur with method names in the future, and to ensure that such intentions are fully met insofar as such classes possessing the needed methods.

This pull request adds an adapter interface (this contributes towards the work currently being done at the `dev@220` branch). This pull request also slightly cleans up some of the PHPDoc comments, and adds both parameter type hints and return types to all methods which belong to the affected classes.

### DIFF
--- a/src/Archive/Adapter/AdapterInterface.php
+++ b/src/Archive/Adapter/AdapterInterface.php
@@ -1,0 +1,42 @@
+<?php
+/**
+ * This file is part of the Zest Framework.
+ *
+ * @author Muhammad Umer Farooq (Malik) <mumerfarooqlablnet01@gmail.com>
+ *
+ * @link https://lablnet.github.io/profile/
+ *
+ * For the full copyright and license information, please view the LICENSE
+ *  file that was distributed with this source code.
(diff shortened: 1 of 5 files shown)


peter279k:
- I wonder why this PR is compared to the `dev@220` branch rather than `master` branch.
- All latest commits are available on `master` branch, not on `dev@220` branch.

Maikuolan:
> - I wonder why this PR is compared to the `dev@220` branch rather than `master` branch.

This is intentional. This PR proposes some changes to the archive files, which currently exist in `dev@220`, but which don't yet exist in `master` (due to still being worked on at #311).

My idea behind targeting `dev@220` is that, if and when this PR is merged, the commits here (#321) will appear at the end of the list of commits at the other PR (#311), but not yet at master (since the archive files are still being worked on). Then everything can appear in master later, after we're ready to merge the other PR. :-)

lablnet (approved):

PR #322, opened 2020-11-25
Fixes, improvements, tests (archive classes).

Maikuolan:
Various fixes and improvements on the archive classes, and an attempt to write some tests for them (more details attached to commit descriptions).

I'm not quite as proficient at writing tests as I would like, so let's wait and see whether these tests pass before merging, in case anything isn't quite right. Fingers crossed.


Maikuolan:
Hm.. Okay.. There are already some complaints from Sider and StyleCI. I'll try to fix those first.

Maikuolan:
> The method compress() has a Cyclomatic Complexity of 21. The configured cyclomatic complexity threshold is 10.

I'm not really sure what would be the best way to fix that last one. It seems, Sider doesn't like the number of if/switch/else statements..? I could reduce the number of such statements, but not without sacrificing readability, I think.

All the other Sider and StyleCI complaints have been fixed though.

Still waiting for the TravisCI results (pending.. pending.. pending.. hmm.. '^.^).

lablnet:
> > The method compress() has a Cyclomatic Complexity of 21. The configured cyclomatic complexity threshold is 10.
> 
> I'm not really sure what would be the best way to fix that last one. It seems, Sider doesn't like the number of if/switch/else statements..? I could reduce the number of such statements, but not without sacrificing readability, I think.

I wondered to see in sider, i think we should remove sider from project, what do you think?

`travis` is too slow, in master branch we already moves to GitHub action

at end, thanks for your time to contribute.

peter279k:
I think it should let `zestframework:dev@220` compare to `master` branch to make current version consistency.

Maikuolan:
> I wondered to see in sider, i think we should remove sider from project, what do you think?

Could be worth considering. I don't use Sider at any other projects, so I don't have much experience with it, but I suspect that worrying too much about some of the complaints from Sider could potentially hinder further progress in the future, depending on the naturally expected complexity of possible future functionality that we may want to see implemented.

> I think it should let `zestframework:dev@220` compare to `master` branch to make current version consistency.

Good point. ^.^

src/Archive/Adapter/Bzip.php
```diff
@@ -27,54 +32,63 @@ class Bzip extends AdapterInterface
...
-     * @return bool
+     * @return bool True when succeeded; False when failed.
      */
-    public function compress(string $file, string = $destination = '', bool $overwrite = false): bool
+    public function compress($files, string $destination = '', bool $overwrite = false): bool
     {
-        //if the zip file already exists and overwrite is false, return false
+        // If the destination already exists and overwrite is false, return false.
         if (file_exists($destination) && !$overwrite) {
             return false;
```
lablnet: As the file already exists, we should return true here, or we have to remove `compressed` file from tests.

src/Archive/Adapter/Gzip.php
```diff
@@ -27,61 +32,68 @@ class Gzip extends AdapterInterface
...
-     * @return bool
+     * @return bool True when succeeded; False when failed.
      */
-    public function compress(string $file, string = $destination = '', bool $overwrite = false): bool
+    public function compress($files, string $destination = '', bool $overwrite = false): bool
     {
-        //if the zip file already exists and overwrite is false, return false
+        // If the destination already exists and overwrite is false, return false.
         if (file_exists($destination) && !$overwrite) {
             return false;
```
lablnet: Same as describe above case.

lablnet (changes requested):

src/Archive/Adapter/Gzip.php
lablnet: Same case with `zip` class

src/Archive/Adapter/Bzip.php
Maikuolan: Ah, good catch.

Maikuolan:
Done. :-)

I guess then, we can see whether the tests pass with GitHub Actions..? ^.^

lablnet:
> I guess then, we can see whether the tests pass with GitHub Actions..? ^.^

image

For now test pssing, and after the PR merge i think we can able to see either tests also working with github actions or not.

lablnet:
This `PR` is ready to merge

lablnet (approved):

PR #325, opened 2020-11-28
console

lablnet:
As i was thinking since the long time about new improved `console` so finally i try to find time and work on it.

Its new console, which support to register commands etc..
A simple hello world command may look like this

```php
<?php

namespace App\Console\Commands;
use Zest\Console\Command;

class Hello extends Command
{
    protected $sign = "hello";
    protected $description = "Command to print hello message on screen";

    public function handle()
    {
        $this->write("<green>Hello World.</green>");
    }
}
```
In `Zest` Skeleton it should added to 
`App/Config/Console.php`
```php
<?php

namespace Config;
use Zest\Console\Commands as InternalCommands;

class Commands extends InternalCommands {
    protected $commands = [
        ['hello',  \App\Console\Commands\Hello::class],
    ];
}


```
It also have few internals commands

Here is snapshot of list command
rsz_screenshot_from_2020-11-28_18-24-34

@peter279k  and @Maikuolan  what do yo think about this approach?

The `symfony console` exists no matter, but i love to reinvent wheel, and the motive of Zest framework.


peter279k:
Your works are very similar with [Symfony/Components](https://symfony.com/doc/current/components/console.html).

I don't think it's not very necessary to rebuild a console....Why don't you build a console on top of above component :)?

lablnet:
> Your works are very similar with [Symfony/Components](https://symfony.com/doc/current/components/console.html).
> 
> I don't think it's not very necessary to rebuild a console....Why don't you build a console on top of above component :)?

Because

>> The main purpose of Zest framework is to provide a very lightweight framework without any external dependencies except core and autoloader

Maikuolan:
Seems like an interesting idea. Would be a nice addition to Zest, I think. :-)

I'm not super familiar with framework-based consoles though, TBH. Even Symfony's console, which seems to be reasonable popular AFAICT, isn't something I tend to ever use, except when necessary (e.g., when using a Symfony-based application which relies on their console).

peter279k:
Just notice that the console about Laravel framework relies on the `Symfony/Console`.

My main suggestion is: We should consider using the lightweight PHP console to build the console Zest framework.

It should be easy to build and it can avoid "repeating yourself" :).

lablnet:
> My main suggestion  is: We should consider using the lightweight PHP console to build the console Zest framework.

What do you mean by light weight php console

@peter279k

peter279k:
You can find the console via the [packagist.org ](https://packagist.org/?query=console) site.

lablnet:
> You can find the console via the [packagist.org ](https://packagist.org/?query=console) site.

For now in this 'PR' I will continue a bit work the think about it. The proglen is that to include third party lib is violate term of Zest .

peter279k:
> > You can find the console via the [packagist.org ](https://packagist.org/?query=console) site.
> 
> For now in this 'PR' I will continue a bit work the think about it. The proglen is that to include third party lib is violate term of Zest .

Agree. Go ahead with that.

src/Console/Output.php
```diff
@@ -17,29 +17,92 @@
...
 
 namespace Zest\Console;
+use Zest\Console\Input\Table;
 
 class Output extends Colorize
 {
 
-    public function write($value, $newLine = false)
+    /**
+     * Queit.
```
Maikuolan: Typo.

lablnet: Fixed :+1:

src/Console/Colorize.php
```diff
@@ -0,0 +1,84 @@
...
+     * Get the color by key.
+     *
+     * @param string $color      Color key
+     * @param bool   $background
+     *
+     * @since 3.0.0
+     *
+     * @var string
+     */
+    public function get($color, $background = false)
```
peter279k: It should be `public function get(string $color, bool $background = false)` due to be consistency.

src/Console/Command.php
```diff
@@ -0,0 +1,135 @@
...
+
+    /**
+     * Function to handle the class.
+     *
+     * @param string $str
+     * @param bool   $newLine
+     *
+     * @return void
+     */
+    public function write($str, $newLine = true)
```
peter279k: It should be `public function write(string $str, bool $newLine = true)` due to be consistency.

src/Console/Command.php
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /**
+     * Prompt user for input.
+     *
+     * @param string $str
+     *
+     * @return void
+     */
+    public function ask($str)
```
peter279k: It should be `public function ask(string $str)` due to be consistency.

src/Console/Command.php
```diff
@@ -0,0 +1,135 @@
...
+
+        return (new Input())->ask();
+    }
+
+    /**
+     * Terminate the console.
+     *
+     * @return void
+     */
+    public function terminate()
```
peter279k: It should be `public function terminate(): void`.

src/Console/Command.php
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /**
+     * Function to handle the class.
+     *
+     * @param \Zest\Console\Output $output
+     *
+     * @return void
+     */
+    abstract public function handle(Output $output);
```
peter279k: It should be `abstract public function handle(Output $output): void;`.

src/Console/Commands/Cache.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Function to handle the class.
+     *
+     * @param \Zest\Console\Output $output
+     *
+     * @return void
+     */
+    public function handle(Output $output)
```
peter279k: It should be `public function handle(Output $output): void`.

src/Console/Commands/Controller.php
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * Function to handle the class.
+     *
+     * @param \Zest\Console\Output $output
+     *
+     * @return void
+     */
+    public function handle(Output $output)
```
peter279k: It should be `public function handle(Output $output): void`.

src/Console/Commands/ListCmd.php
```diff
@@ -0,0 +1,102 @@
...
+        parent::__construct();
+        $this->container = new Container();
+    }
+
+    /**
+     * Get the list of commandss.
+     *
+     * @return void
+     */
+    public function getList()
```
peter279k: It should be `public function getList(): void`.

src/Console/Commands/ListCmd.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Function to handle the class.
+     *
+     * @param \Zest\Console\Output $output
+     *
+     * @return void
+     */
+    public function handle(Output $output)
```
peter279k: It should be `public function handle(Output $output): void`.

src/Console/Commands/Version.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * Function to handle the class.
+     *
+     * @param \Zest\Console\Output $output
+     *
+     * @return void
+     */
+    public function handle(Output $output)
```
peter279k: It should be `public function handle(Output $output): void`.

src/Console/Input/Table.php
```diff
@@ -0,0 +1,123 @@
...
+        $this->header = $header;
+        $this->items = $items;
+    }
+
+    /**
+     * Print table border.
+     *
+     * @since 3.0.0
+     *
+     * @return array
```
peter279k, line 60: Wrong comment annotation about returning type.

It should be `@return string|self`.

src/Console/Input/Table.php
```diff
@@ -0,0 +1,123 @@
...
+     */
+    public function printBorder($row): self
```
peter279k, line 62: The native type `self` hint should be removed because it will return `string` or `self` type hints.

src/Console/Output.php
```diff
@@ -0,0 +1,111 @@
...
+    private $quiet = false;
+
+    /**
+     * Mark the console quiet.
+     *
+     * @since 3.0.0
+     *
+     * @return self
+     */
+    public function quiet()
```
peter279k: It should be `public function quiet(): self`.

src/Console/Output.php
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * Exit.
+     *
+     * @since 3.0.0
+     *
+     * @return void
+     */
+    public function exit()
```
peter279k: It should be ` public function exit(): void`.

lablnet:
## TODO
- [x] Output List(s)
- [x]  Parsing arguments and flags.
- [x] Support Args and flag in commands.


After that this `PR`will ready to merge.

PR #326, opened 2020-11-28
Adopt null coalescence.

Maikuolan:
Since we've set the minimum PHP requirement for Zest to a more modern PHP version, we're able to take advantage of PHP's null coalescence operator, which in some cases can replace existing isset checks, meaning less code.

### DIFF
--- a/src/http/Request.php
+++ b/src/http/Request.php
@@ -28,13 +28,13 @@ class Request extends Uri
     public function __construct()
     {
         parent::__construct();
-        $this->get = (isset($_GET)) ? $_GET : [];
-        $this->post = (isset($_POST)) ? $_POST : [];
-        $this->files = (isset($_FILES)) ? $_FILES : [];
-        $this->session = (isset($_SESSION)) ? $_SESSION : [];
-        $this->cookie = (isset($_COOKIE)) ? $_COOKIE : [];
-        $this->server = (isset($_SERVER)) ? $_SERVER : [];
-        $this->env = (isset($_ENV)) ? $_ENV : [];
+        $this->get = $_GET ?? [];
+        $this->post = $_POST ?? [];
+        $this->files = $_FILES ?? [];
+        $this->session = $_SESSION ?? [];
+        $this->cookie = $_COOKIE ?? [];
+        $this->server = $_SERVER ?? [];
+        $this->env = $_ENV ?? [];
 
         if ($this->getRequestMethod()) {
             $this->parseData();
@@ -184,7 +184,7 @@ class Request extends Uri
      */
     public function isSecure()
     {
-        return ($this->gethttps() || $this->getServerPort() && $this->getServerPort() == '443') ? true : false;
+        return ($this->gethttps() || $this->getServerPort() && $this->getServerPort() == '443');
     }
 
     /**
--- a/src/http/Request.php
+++ b/src/http/Request.php
@@ -184,7 +184,7 @@ class Request extends Uri
      */
     public function isSecure()
     {
-        return ($this->gethttps() || $this->getServerPort() && $this->getServerPort() == '443');
+        return $this->gethttps() || $this->getServerPort() && $this->getServerPort() == '443';
     }
 
     /**
--- a/src/http/HTTP.php
+++ b/src/http/HTTP.php
@@ -295,7 +295,7 @@ abstract class HTTP extends Headers
      */
     public function getRequestMethod()
     {
-        return (isset($_SERVER['REQUEST_METHOD'])) ? $_SERVER['REQUEST_METHOD'] : false;
+        return $_SERVER['REQUEST_METHOD'] ?? false;
     }
 
     /**
@@ -307,7 +307,7 @@ abstract class HTTP extends Headers
      */
     public function getRequestUrl()
     {
-        return (isset($_SERVER['REQUEST_URI'])) ? $_SERVER['REQUEST_URI'] : false;
+        return $_SERVER['REQUEST_URI'] ?? false;
     }
 
     /**
@@ -319,7 +319,7 @@ abstract class HTTP extends Headers
      */
     public function getContentType()
     {
-        return (isset($_SERVER['CONTENT_TYPE'])) ? $_SERVER['CONTENT_TYPE'] : false;
+        return $_SERVER['CONTENT_TYPE'] ?? false;
     }
 
     /**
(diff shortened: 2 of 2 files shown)


Maikuolan:
(StyleCI complaint. One moment. I'll fix that now).

lablnet (approved):

Maikuolan:
Fixed. Scrutinizer taking a long time.. Not sure what's happening there. PHP 7.2 workflow failure outside the scope of this PR (originates with [`1058999`](https://github.com/zestframework/Zest_Framework/commit/10589995c2bddbbfe706bc6b1dec0344d65e9916)). Should be okay fine. :+1:

lablnet:
> Fixed. Scrutinizer taking a long time.. Not sure what's happening there. PHP 7.2 workflow failure outside the scope of this PR (originates with [`1058999`](10589995c2bddbbfe706bc6b1dec0344d65e9916)). Should be okay fine.

Yes about that i think latest phpunit is not supported by php 7.2!

Maikuolan:
> Yes about that i think latest phpunit is not supported by php 7.2!

Yep. :-)

The problem is with PHPUnit 9.4. PHPUnit 9.4 requires at least PHP >= 7.3 (PHPUnit 8 seems to still support PHP 7.2, but I guess, with PHP 8.0 now available, we're forced onto the newer PHPUnit version. It's a somewhat difficult problem).

lablnet:
> Fixed. Scrutinizer taking a long time.. Not sure what's happening there. PHP 7.2 workflow failure outside the scope of this PR (originates with [`1058999`](https://github.com/zestframework/Zest_Framework/commit/10589995c2bddbbfe706bc6b1dec0344d65e9916)). Should be okay fine.


```sh
PHPUnit 9.4.3 by Sebastian Bergmann and contributors.

....FF........................................................... 65 / 94 ( 69%)
.........SS.SS...............                                     94 / 94 (100%)

Time: 00:00.966, Memory: 22.00 MB

There were 2 failures:

1) Framework\Tests\ZipTest::testExtract

/home/scrutinizer/build/Tests/Archive/ZipTest.php:19

2) Framework\Tests\ZipTest::testCompress

/home/scrutinizer/build/Tests/Archive/ZipTest.php:27

FAILURES!
Tests: 94, Assertions: 171, Failures: 2, Skipped: 4.

Generating code coverage report in Clover XML format ... done [00:02.023]
```
Scrutinizer seems failing the `Archieve` Test, today morning GMT+5 i test on my Ubuntu all of them tests were fine, its consufing!!!!!!!!!!!!!!!1

lablnet:
> The problem is with PHPUnit 9.4. PHPUnit 9.4 requires at least PHP >= 7.3 (PHPUnit 8 seems to still support PHP 7.2, but I guess, with PHP 8.0 now available, we're forced onto the newer PHPUnit version. It's a somewhat difficult problem).

if  we say hey, we required php 7.3 does it will be bad idea? what do you think.

Maikuolan:
> if we say hey, we required php 7.3 does it will be bad idea? what do you think.

It is normal, I think, that well-maintained PHP packages eventually end up increasing their minimum required PHP version over time, as they introduce newer features, update syntax, refactor their codebase, etc. There's nothing wrong with doing that. However, IMHO, increasing the minimum required PHP version should *always* be regarded as a significant backwards-incompatible change (in the sense that backwards compatibility with older PHP versions will be affected as a result of increasing the minimum required PHP version, which could break installations and cause problems for some users if they aren't properly prepared for it).

Since it is good practice in modern PHP packages to also follow semantic versioning (semver), and there's a rule in semver that states that whenever implementing significant backwards-incompatible changes, that the major version number of the package should increase, I've written a specific rule within the contribution guidelines for some of my own packages (e.g., CIDRAM, phpMussel, Common Classes Package, etc) that the minimum PHP version required by the package should *only* ever be increased when a new major version is released.

Since the current latest version of Zest (according to repository tags) is v2.0.3, if Zest were to follow a similar set of rules or ideas, then the next version of Zest to be released should be v3.0.0.

Some people may consider it a little early for that right now, seeing as the current *minor* version of Zest is __0__ (the 0 in v2.__0__.3), but if there's a legitimate reason for proceeding, then that shouldn't really matter much. But, that may be something to think about, maybe.

We could increase the minimum required PHP version immediately, or.. if you would prefer to schedule 3.0.0 at a specific, later date in the future.. maybe we could delay increasing the minimum required PHP version until a little closer to that date, planning ahead for it to occur for that future release..? There are several approaches we could take here, each with their pros and cons, worth thinking about a little.

Maikuolan:
Also, do we have any demographic information about Zest's userbase? Things like which PHP versions Zest's users are using? A little difficult to acquire that information, but if we're able to acquire it, and if we then have some idea about what proportion of Zest's users are actually using PHP 7.2, we may then have a stronger idea about whether increasing the minimum required PHP version would have much impact (and of course, nobody should be using anything older than PHP 7.2, since the minimum required PHP version is already at 7.2 currently).

lablnet:
> We could increase the minimum required PHP version immediately, or.. if you would prefer to schedule 3.0.0 at a specific, later date in the future.. maybe we could delay increasing the minimum required PHP version until a little closer to that date, planning ahead for it to occur for that future release..? There are several approaches we could take here, each with their pros and cons, worth thinking about a little.

Ummm, yet I am not sure but I expect for release date of 3.0.0 may be April 2021,the reason why I am not sure is that am student and have tto study and other works etc.. so

Also we need to make it stable with more features like group routing, console. Etc.

And thanks for your explanation I am going to look your packages Contribution guide to understand it more deeply.

lablnet:
> Also, do we have any demographic information about Zest's userbase? 

Yes it's difficult and have no info, but at this stage I can confidently say no one yet using it.

That why I am back to Zest to make it stable as much as I can so people like to use it.

After 3.0.0 I will benchmark it using PHP benchmark kit  then I hope it will be fastest then others so alternatively some people like to use it.

src/http/Request.php
```diff
@@ -184,7 +184,7 @@ public function isPatch()
      */
     public function isSecure()
     {
-        return ($this->gethttps() || $this->getServerPort() && $this->getServerPort() == '443') ? true : false;
+        return $this->gethttps() || $this->getServerPort() && $this->getServerPort() == '443';
```
peter279k, line 187: According to the official [PHP doc](https://www.php.net/manual/en/reserved.variables.server.php) about `$_SERVER['SERVER_PORT']`, the port number type is string.

For value comparing performance, it should use the `===` to make value equals strictly.

lablnet: :+1:

PR #327, opened 2020-11-28
Experimental

lablnet:


peter279k:
Are you planning to drop PHP 7.2 version support?

lablnet:
> Are you planning to drop PHP 7.2 version support?

Yes, because the reason behind is that the latest phpunit version also dropped it. and i love to take advantage of latest php version but till the version 5 of Zest we should support from 7.3 and above, and after that we may say oh now we need php 8 required version it can be after two to three years.

PR #340, opened 2020-12-05
Update package references in README.

Maikuolan:
Replaces `softhub99/zest_framework` (marked as abandoned at Packagist) with `zest/framework` (Zest's current package identity).

PR #341, opened 2020-12-06
Attempt to fix Gzip failures.

Maikuolan:
(See comments at #340).

### DIFF
--- a/src/Archive/Adapter/Gzip.php
+++ b/src/Archive/Adapter/Gzip.php
@@ -84,7 +84,7 @@ class Gzip implements AdapterInterface
         }
 
         $filename = $destination;
-        if ($outfile = gzopen($filename, 'wb' . $mode)) {
+        if ($outfile = gzopen($filename, 'wb'.$mode)) {
             if ($infile = fopen($files, 'rb')) {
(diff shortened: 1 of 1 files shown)


Maikuolan:
StyleCI complaints. '-.-

One sec.

Maikuolan:
Tests look good now. :-)

lablnet (approved):
👍

src/Archive/Adapter/Gzip.php
```diff
@@ -84,7 +84,7 @@ public function compress($files, string $destination = '', bool $overwrite = fal
         }
 
         $filename = $destination;
-        if ($outfile = gzopen($filename, $mode)) {
+        if ($outfile = gzopen($filename, 'wb'.$mode)) {
```
peter279k, line 87: I think  it should verify the compression mode before calling the `gzopen` function.

According to the [gzcompress](https://www.php.net/manual/en/function.gzcompress.php) reference, the compression level range should be from `0` to `9`.

lablnet: > I think it should verify the compression mode before calling the `gzopen` function.

agreed :+1:

